PR #11, opened 2016-07-12
disable execstack explicitly for libxorgxrdp

kenhys:
rpmlint reports a warning about execstack.

  xorg-x11-drv-xrdp.x86_64: W: executable-stack
  /usr/lib64/xorg/modules/libxorgxrdp.so The binary declares the stack
  as executable.  Executable stack is usually an error as it is only
  needed if the code contains GCC trampolines or similar constructs
  which uses code on the stack.  One common source for needlessly
  executable stack cases are object files built from assembler files
  which don't define a proper .note.GNU-stack section.

It is better to disable needless executable stack.

### DIFF
--- a/module/Makefile.am
+++ b/module/Makefile.am
@@ -32,7 +32,7 @@ AM_CFLAGS = \
 
 libxorgxrdp_la_LTLIBRARIES = libxorgxrdp.la
 
-libxorgxrdp_la_LDFLAGS = -module -avoid-version
+libxorgxrdp_la_LDFLAGS = -module -avoid-version -Wl,-z,noexecstack
 
 libxorgxrdp_ladir = $(moduledir)
 


proski:
I don't like adding linker flags unconditionally. Those flags may not be portable. The reason for the fix doesn't seem to be sufficient. If rpmlint doesn't like something, the first approach should be to fix the packaging instructions (the specfile). Contacting upstream is reasonable if the code is doing something wrong or doesn't allow flexibility needed for the packaging.

Why is rpmlint complaining of xorgxrdp, but not of other packages? I don't think every other package passes linker flags to disable executable stack.

In particular, please see what other X11 modules are doing differently. Do they respect some flags that xorgxrdp ignores? Then that should be in the fix.

metalefty:
A better fix has been introduced in #14.  Not to merge this accidentally, I close this.

PR #14, opened 2016-10-26
Mark stack non-executable in *.asm files

proski:
This is a more portable alternative to #11 

The fix is described at https://wiki.gentoo.org/wiki/Hardened/GNU_stack_quickstart

The fix was tested by running the command that rpmlint runs: `readelf -W -S -l -d -s libxorgxrdp.so` and checking that "E" is not in the flag column for the GNU_STACK headers.


metalefty:
Debian guys have quite similar patch to this.  Debian's xorgxrdp is patched and works with no problem.
Meaning, this looks good to me.

proski:
I've updated patch. The new version takes care not to break builds for non-ELF systems, such as Mac OS.

PR #16, opened 2016-10-28
Fix permission issues with sesman.

Natureshadow:
Accompanied by a patch to xrdp.

### DIFF
--- a/module/rdpClientCon.c
+++ b/module/rdpClientCon.c
@@ -1127,7 +1127,7 @@ rdpClientConInit(rdpPtr dev)
                 return 0;
             }
         }
-        g_chmod_hex("/tmp/.xrdp", 0x1777);
+        g_chmod_hex("/tmp/.xrdp", 0x3777);
     }
     i = atoi(display);
     if (i < 1)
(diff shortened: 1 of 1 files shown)


module/rdpClientCon.c
```diff
@@ -1127,7 +1127,7 @@ rdpClientConInit(rdpPtr dev)
                 return 0;
             }
         }
-        g_chmod_hex("/tmp/.xrdp", 0x1777);
+        g_chmod_hex("/tmp/.xrdp", 0x3777);
```
proski: Anything that touches suid and sgid bits should be documented. Also, it's it highly desirable that such behavior can be configured using configuration files. Creating a directory with a predictable name in a world writable directory was bad already. Creating a sgid directory makes it even worse. What if an attacker replaces /tmp/.xrdp with a link to an executable and exploits this code to get sgid permissions? What if the process group is root or wheel?

module/rdpClientCon.c
```diff
@@ -1145,7 +1145,13 @@ rdpClientConInit(rdpPtr dev)
             LLOGLN(0, ("rdpClientConInit: g_tcp_local_bind failed"));
             return 1;
         }
-        g_sck_listen(dev->listen_sck);
+        if (g_sck_listen(dev->listen_sck) != 0)
```
proski: This looks like an unrelated change. Please submit unrelated changes as separate commits, even if they are in the same pull request.

metalefty:
Do not just submit downstream patches to upstream.  If a downstream patch contains fixes for more than one issue, please separate the patch and create PRs for each patch.

PR #13, opened 2016-10-24
Fixing "make dist" and other build system issues

proski:

PR #17, opened 2016-10-28
Miscelleanous fixes

Natureshadow:
Accompanies https://github.com/neutrinolabs/xrdp/pull/467

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -9,7 +9,7 @@ AC_PROG_CC
 AC_C_CONST
 AC_PROG_LIBTOOL
 
-AM_CONDITIONAL(GOT_PREFIX, test "x${prefix}" != "xNONE"])
+AM_CONDITIONAL(GOT_PREFIX, [test "x${prefix}" != "xNONE" && test "x${prefix}" != "x/usr"])
 
 AC_CHECK_HEADER([xorg/xorg-server.h], [],
   [AC_MSG_ERROR([please install xserver-xorg-dev or xorg-x11-server-sdk])])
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -9,7 +9,7 @@ AC_PROG_CC
 AC_C_CONST
 AC_PROG_LIBTOOL
 
-AM_CONDITIONAL(GOT_PREFIX, test "x${prefix}" != "xNONE"])
+AM_CONDITIONAL(GOT_PREFIX, [test "x${prefix}" != "xNONE" && test "x${prefix}" != "x/usr"])
```
proski: GOT_PREFIX is not needed at all. My pull request #13 removes it. The problem with GOT_PREFIX is that it doesn't have a clear semantic - when it should be defined and what it should do. The implication is that if the user installs software to some location, includes and libraries already in that location should be used. The problem is that it's not always the case, and it's easier to add a directory to the includes than to remove it.

xrdpkeyb/rdpKeyboard.c
```diff
@@ -668,7 +670,8 @@ rdpkeybControl(DeviceIntPtr device, int what)
     switch (what)
     {
         case DEVICE_INIT:
-            rdpkeybDeviceInit(device, &keySyms, modMap);
+            if (rdpkeybDeviceInit(device, &keySyms, modMap))
+                return BadAlloc;
```
proski: I don't like returning BadAlloc here. It created a tight coupling. One function knows why another function fails. Perhaps rdpkeybDeviceInit() itself could return BadAlloc, and its callers could pass it further up.

proski:
The PR can be closed now. GOT_PREFIX has already been removed. Delaying exit on memory allocation failure is wrong in my opinion, and we haven't heard from the PR author.

PR #18, opened 2016-11-07
Add support for new API in RegisterBlockAndWakeupHandlers()

proski:
ServerBlockHandlerProcPtr was introduced for the new signature of the block handler. Changes to the wakeup handler were made in the came commit. Test if type ServerBlockHandlerProcPtr is defined. If yes, use the new API.


jsorg71:
I prefer to use XORG_VERSION_NUMERIC for things like this.
I'll make a change to it.

jsorg71:
merged as b4e0452

proski:
No problem. I didn't want to figure out the cutoff version.

PR #20, opened 2016-11-10
Compilation issue: typo fix in rdpClientConAddEnabledDevice name

proski:
One small fixup for the latest patches. With it, xorgxrdp compiles without warnings and works on Fedora 25.


jsorg71:
merged as f415724

Issue #22, opened 2016-11-12
DPI setting is ignored, always using ~100 dpi


proski:
No matter what DPI I set on the Xorg command line with the `-dpi` argument, the xorgxrdp window shows the same font size. It is slightly bigger than what other backends (X11rdp and Xvnc) show with 96 dpi. At 100 dpi, the font sizes are the same. At 144 dpi, other backends show much larger fonts, as can be seen in the screenshot.
![screen shot 2016-11-11 at 6 39 31 pm](https://cloud.githubusercontent.com/assets/1317472/20235331/d7cc5a14-a843-11e6-9f9b-cf7c01f307af.png)

jsorg71:
Possibly fixed in devel

proski:
That won't work. monitorResolution is 0 if `-dpi` is not specified on the command line. That would lead to division by zero. I made another patch, I'll submit it shortly. It may be a bit overengineered, but it actually respects horizontal and vertical DPI.

proski:
Fixed now, resolving.

PR #24, opened 2016-11-12
Use xDpi and yDpi instead of monitorResolution

proski:
monitorResolution is 0 if DPI is not specified on the Xorg command line, so it should not be used without checking.

This patch respects the horizontal and the vertical DPI in case they are different.


proski:
This actually fixes the issue #22. Tested with and without `-dpi` parameter.

PR #30, opened 2016-11-17
Miscellaneous fixes for the next release

proski:
There are no functional changes. The main focus is on high standards for the build system.

The most important patch is changing the package name to xorgxrdp :)

Enabling -Wwrite-string will introduce some unfixed warnings. I believe they are harmless. I'll fix them separately.

nasm is still not as silent as I would like it to be. The remaining message comes from `nasm_lt.sh`.

The change in readme.txt is just to note the change. That file needs to be rewritten.


jsorg71:
Thanks, I'll review and manually merge.

proski:
I've adjusted the xrdp header check so it doesn't look in  `..` as xorgxrdp is not a submodule of xrdp anymore.

proski:
The test is non-interactive now. It can be run by `make check`, which allows us to integrate it with travis.

proski:
I rebased everything. This branch has the changes that were not on it yesterday. The descriptions are correct, they reflect the remaining changes.

jsorg71:
Ok, great, these look at as is.

jsorg71:
I'm getting a fail on make check is DISPLAY is set before

jsorg71:
I changed it to TEST_DISPLAY

proski:
Thank you, that's the right approach.

PR #35, opened 2016-11-19
Reduce dependency on xrdp to simplify compilation

proski:


proski:
I don't want to compile or even configure xrdp just to have one header. This is especially useful for continuous integration. Instead of checking out xrdp, we could just download the required header.

PR #38, opened 2016-11-23
#endif typo on amd64

limadm:
This small typo is causing compilation problems on amd64_sse2.


itamarjp:
@proski please review

proski:
It's fixed in the `devel` branch. I think it's time to merge it into `master`. No need to create separate patches for the same issue.

jsorg71:
I think I merged devel to master

metalefty:
Already fixed in devel and devel has been merged into master, closing.

PR #39, opened 2016-11-24
Add COPYING file

proski:
The license text is the same as in the xorgxrdp sources. "make dist" includes COPYING file automatically if it's present.


metalefty:
No copyright year and copyright holders however they can be added later. Merging.

PR #41, opened 2016-12-01
Improve cleanup code

proski:
This PR improves the cleanup when Xorg exits. Changes are mostly for code quality. A few hacks have been removed. The screen-related cleanup doesn't depend on the cursor or input code anymore. Valgrind won't complain of leaked file descriptors.

The teardown function is not called for non-input devices. That's still the case. The xorgxrdp module has a function `xorgxrdpTearDown` that is not called. Hopefully we fix teardown one day, and then it will be called. Another function, `xorgxrdpDownDown` is called manually to do the cleanup.

`xorgxrdpDownDown` was called from the cursor code. It is strange and unreliable. Older RHEL 6 based systems don't call that function.

Instead, `xorgxrdpDownDown` should be called from `rdpCloseScreen`. It turns out that `rdpCloseScreen` was not properly registered (no code was using `rdpCloseScreen` except `rdpCloseScreen` itself), so the let's fix it.

Now that `xorgxrdpDownDown` is working reliably, we can remove a hack that removes `/tmp/.xrdp/xrdp_display_$DISPLAY` when the keyboard device is unregistered. That file (actually, UNIX socket) is already removed in the cleanup code, I just needed to add logging.

Finally, Valgrind complains that a file descriptor was left open. To fix it, close the client connection socket (if it's open) in addition to the listening socket.

### DIFF
--- a/module/rdpClientCon.c
+++ b/module/rdpClientCon.c
@@ -1205,6 +1205,15 @@ int
 rdpClientConDeinit(rdpPtr dev)
 {
     LLOGLN(0, ("rdpClientConDeinit:"));
+
+    if (dev->clientConTail != NULL)
+    {
+        LLOGLN(0, ("rdpClientConDeinit: disconnecting only clientCon"));
+        rdpClientConDisconnect(dev, dev->clientConTail);
+        dev->clientConHead = NULL;
+        dev->clientConTail = NULL;
+    }
+
     if (dev->listen_sck != 0)
     {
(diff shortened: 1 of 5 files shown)


metalefty:
LGTM

jsorg71:
Thanks @proski 
Much better then that hack we had.  It's really important that the cleanup runs.  Without it, we would run out of available sessions.

proski:
Thanks. Let's merge this PR before the release.

PR #42, opened 2016-12-02
Provide reasonable hTotal, vTotal and dotClock for the display mode

proski:
gtk3 programs crash when they divide dotClock by hTotal and vTotal to
calculate the refresh rate.

That is not an issue if Xorg supports RandR 1.5, but may be an issue for
older versions of Xorg.


speidy:
+1

PR #45, opened 2016-12-05
Convert README to Markdown

metalefty:


proski:
Two things.

Please keep the document readable as text, say, in less. It's possible to wrap long lines and have more generous vertical spacing without affecting the markdown output.

Secondly, the change breaks `make dist`. I have an idea how to fix `make distcheck`, I'll submit it shortly together with the Travis rule change.

PR #46, opened 2016-12-05
Add copyright year and holder

### LINKED ISSUE #37: Need a license file
xorgxrdp should not be released without a license. The source files appear to carry some variant of the X.Org license.

https://www.x.org/releases/X11R7.7/doc/xorg-docs/License.html

But ultimately it should be up to the authors of the code.

The common convention is to put the license to a file called COPYING in the top source directory.

metalefty:
Closes #37

### DIFF
--- a/COPYING
+++ b/COPYING
@@ -1,3 +1,8 @@
+(The X11 License)
+
+Copyright (C) 2014-2016, Jay Sorg <jsorg71@gmail.com>
+All Rights Reserved.
+
 Permission to use, copy, modify, distribute, and sell this software and its
 documentation for any purpose is hereby granted without fee, provided that
 the above copyright notice appear in all copies and that both that


metalefty:
Is there any code brought from other software to xorgxrdp?
If no, let's merge this.

proski:
It looks like it's actually The Open Group license, but without the last line.
https://www.x.org/releases/X11R7.7/doc/xorg-docs/License.html#X_MIT_Copyrights

Perhaps the license name should be after the author's name, not before?

metalefty:
The license is the [MIT License](https://opensource.org/licenses/MIT). I believe it is also called the X11 License[1].
Not to mislead, I call it "The X11 License".



The license name can be placed wherever because it is completely unnecessary. The reason why I added it is to clarify. The X11 License is very similar to other licenses such as the ISC License, the modified BSD License. The X11 license and the modified BSD license are almost equivalent[2].


[1] http://whatis.techtarget.com/definition/MIT-License-X11-license-or-MIT-X-license
> The MIT License (also known as the X11 license or MIT X license) is a software license that was originally developed at the Massachusetts Institute of Technology.  

[2] https://www.gnu.org/licenses/license-list.en.html#ModifiedBSD
> The X11 license and the modified BSD license are more or less equivalent.

PR #49, opened 2016-12-06
Use absolute path in PKG_CONFIG_PATH to fix "make distcheck"

proski:
"make distcheck" runs in a subdirectory, so it could not find xrdp headers using relative path.


itamarjp:
I am not sure if its possible to use  $(pwd) instead.

PR #53, opened 2016-12-07
Don't use "--tag NASM" with libtool

proski:
It makes libtool show an error message. No released version of libtool
has any reference to the NASM tag.


metalefty:
+1

PR #55, opened 2016-12-10
Disable dead code

proski:
I wasted some time today trying to improve code in xorgxrdp that is not actually called. To prevent such waste of time in the future, I decided to find all dead code and comment if out with `#if 0` - `#endif`, Good editors will show such code as disabled.

Maybe somebody with a good knowledge of the code could remove the code that is not going to be used and leave the code that could still be useful. Disabling the code is the first step in that direction.

I also made functions without declarations static, so that it can be easily detected if they ever become unnecessary.

### DIFF
--- a/module/rdpCapture.c
+++ b/module/rdpCapture.c
@@ -262,7 +262,7 @@ rdpCopyBox_a8r8g8b8_to_a8b8g8r8(rdpClientCon *clientCon,
 }
 
 /******************************************************************************/
-int
+static int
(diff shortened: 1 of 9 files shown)


jsorg71:
These function will be used after I get the orders drawing merged from X11rdp after this release.

proski:
Maybe you can check which code is truly dead and which is just waiting for its hour? Some code in `rdpClientConConvertPixels` looks quite ineffective to me, I hope it won't be used in any critical path.

I'm closing this PR and leaving the branch as a note to myself.

PR #56, opened 2017-01-04
Fix cppcheck warnings, disallow warnings in Travis CI

proski:
Eliminate all void pointers in rdpCapture.c, use char or const char pointers as needed.


metalefty:
+1

PR #57, opened 2017-01-10
Correctly silence nasm compilation

proski:
Remove output from nasm_lt.sh, print "NASM" and target name when silent
rules are enabled.


jsorg71:
+1

PR #59, opened 2017-01-15
Stop using g_free(), remove g_malloc() and g_free()

proski:
Xorg uses malloc() and free() without any wrappers, and so should
xorgxrdp.

g_free() is equivalent to free(). C standard requires free() to take no
action on NULL argument.


metalefty:
+1

jsorg71:
merged as 28cdffb2067f41c87c7f0738d8f12e31edad5478

PR #61, opened 2017-01-17
Various cleanups

proski:


metalefty:
+1

jsorg71:
merged as 87313cbc0794e7fc7654693f42a93ae8caf1d995
I didn't merge "Remove unused fields from struct _rdpRec"
I will be merging code from X11rdp soon that will use these fields.

proski:
Thanks. I see `do_dirty_os` in X11rdp. But `disconnectScheduled` is clearly a duplicate of `disconnect_scheduled`, and X11rdp has only one similar variable, `g_disconnect_scheduled`. Would you apply a patch that only removes `disconnectScheduled`?

jsorg71:
You are right.  I removed it with 94d80e7c7c6f4b22986cc7abc8db63e3561c5540
Thanks

Issue #62, opened 2017-01-21
segfault and disconnect on debian server


cypa:
Got X-server segfault and so client disconnect on both windows client and linux client, see logs below.
xorgxrdp and xrdp was compiled from latest source tarball on Debian 8.6.

Connect using windows 10 standart client 
and ubuntu 14.04 standart xfreerdp 1.0.2 with command 
`xfreerdp -f -u vasya -p skejlsuhd 46.4.NN.NN:3398 -z -x b -K Russian`

[xfreerdp1.xorgxrdp.13.log.txt](https://github.com/neutrinolabs/xorgxrdp/files/721025/xfreerdp1.xorgxrdp.13.log.txt)
[xfreerdp1.xorgxrdp.11.log.txt](https://github.com/neutrinolabs/xorgxrdp/files/721027/xfreerdp1.xorgxrdp.11.log.txt)
[win10.xorgxrdp.15.log.txt](https://github.com/neutrinolabs/xorgxrdp/files/721026/win10.xorgxrdp.15.log.txt)

cypa:
no ideas?

metalefty:
I also hit the same issue #9.

metalefty:
can you reproduce segfault again with #64 patch? the log becomes more informative.

cypa:
frankly, after apllying the #64 patch no segfaults occured so far

metalefty:
sure. It isn't a fix.

cypa:
yes, it is still there: no more segfaults and current RDP session is going for 20 hours and continues to go ...
all I did:
- apllied #64 patch manually in vi
- make clean
- make
- sudo make install
that's it!

metalefty:
Please show us the log when you reproduce segfaults.  I also hardly reproduce segfaults.

cypa:
I've got no segfaults since #64 patch applied
logs with segfaults attached to the top comment - that's all I have

cypa:
this is the log of current alive RDP session
[xfreerdp.xorgxrdp.17.log.zip](https://github.com/neutrinolabs/xorgxrdp/files/726113/xfreerdp.xorgxrdp.17.log.zip)

metalefty:
Yes, when you reproduce segfaults.

moobyfr:
The log for xfreerdp1.xorgxrdp.13.log.txt is showing

[249760.685] rdpClientConSendMsg: overrun error len 33504 count 2

[...]

PR #64, opened 2017-01-23
log stream size not only length and connection count

metalefty:
I believe it makes it easier to debug buffer overrun #9, #62.

### DIFF
--- a/module/rdpClientCon.c
+++ b/module/rdpClientCon.c
@@ -396,8 +396,9 @@ rdpClientConSendMsg(rdpPtr dev, rdpClientCon *clientCon)
 
         if (len > s->size)
         {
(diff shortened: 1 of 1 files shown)


module/rdpClientCon.c
```diff
@@ -394,8 +394,9 @@ rdpClientConSendMsg(rdpPtr dev, rdpClientCon *clientCon)
 
         if (len > s->size)
         {
-            LLOGLN(0, ("rdpClientConSendMsg: overrun error len %d count %d",
-                   len, clientCon->count));
+            LLOGLN(0, ("rdpClientConSendMsg: overrun error len %d "
+                       "stream size %d client count %d",
```
proski: Please add commas after the first and the second `%d`

metalefty:
Let's merge this and keep our eyes on buffer overrun issue.

PR #68, opened 2017-01-26
PIC-safe assembly for xorgxrdp

mirabilos:
See also: https://github.com/neutrinolabs/xorgxrdp/issues/67

I believe the code to be correct, as the transformations were mechanical (even if done manually), and this compiles on Debian GNU/Linux on i386 just fine (Hurd and GNU/kFreeBSD are expected to follow).


proski:
I made a branch merging #75 , #73 and this PR. `yuv2rgb_speed` is broken by the changes made in this PR.

mirabilos:
Thank you, it is indeed broken, I see why, and I’m working on a fix.

mirabilos:
@proski this new push should address both your legibility concerns and the crash issue (I tested on an actual i386 system)

module/x86/a8r8g8b8_to_a8b8g8r8_box_x86_sse2.asm
```diff
@@ -22,7 +23,30 @@
...
+%macro get_GOT 0
+	call ..@get_GOT
+%%getgot:
+	add ebx, _GLOBAL_OFFSET_TABLE_ + $$ - %%getgot wrt ..gotpc
+%endmacro
+%endif
+%else
+; not ELF
+%ifdef PIC
+%error Position-Independent Code is currently only supported for ELF
```
proski: Please quote error messages.

mirabilos: Even the nasm manual is inconsistent in whether they quote them, but if it pleases your eyes… sure.

proski:
I was able to test the code on a 32-bit systems. The image is good. The test passes (after rebasing this PR onto the current devel branch).

But I would prefer to have constant data in the `.text` section. It may be inelegant, but it's safe (the data is write protected), portable (support for non-ELF systems is retained), and simple (no GOT).

mirabilos:
I agree, and I will change it once I get more than a few minutes free time… :| I was ill for a while, and work piled up on the dayjob.

mirabilos:
See https://github.com/neutrinolabs/librfxcodec/pull/17 for ongoing discussion; these two PRs are linked.

PR #69, opened 2017-01-27
Suppress useless debug messages from rdpRRGetPanning

proski:
Lower debug level to 10. Print pointers that are actually used by the
function. Don't print crtc, it's not used.


metalefty:
+1

PR #72, opened 2017-02-03
add config header, update copyright year

jsorg71:


mirabilos:
- Who defines `HAVE_CONFIG_H`?
- That macro is not correctly named anyway, `HAVE_CONFIG_AC_H` would be more appropriate
- GNU autoconf upstream recommended IIRC to just unconditionally include it

mirabilos:
Oh nevermind, `AC_CONFIG_HEADERS` defines `HAVE_CONFIG_H` (the extra guard is still likely not necessary).

PR #73, opened 2017-02-04
Determine 64-bit and 32-bit systems by the pointer size in the compiler

proski:
This makes it possible to compile 32-bit assembly by using "-m32" flag to
gcc.

Don't call unknown file format "ELF ?", call it "unknown".

Don't define WITH_SIMD, it's never used.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -38,27 +38,29 @@ AC_ARG_WITH([simd],
     AC_HELP_STRING([--without-simd],[Omit SIMD extensions.]))
 if test "x${with_simd}" != "xno"; then
   # Check if we're on a supported CPU
-  AC_MSG_CHECKING([if we have SIMD optimisations for cpu type])
   case "$host_cpu" in
-    x86_64 | amd64)
-      AC_MSG_RESULT([yes (x86_64)])
+    x86_64 | amd64 | i*86 | x86 | ia32)
       AC_PROG_NASM
-      simd_arch=x86_64
-    ;;
-    i*86 | x86 | ia32)
-      AC_MSG_RESULT([yes (i386)])
-      AC_PROG_NASM
-      simd_arch=i386
-    ;;
+      case "$ptr_size" in
+        8)
+          simd_arch=x86_64
+          AC_MSG_NOTICE([Using x86_64 SIMD instructions])
+          AC_DEFINE([RFX_USE_ACCEL_AMD64], [1], [Use x86_64 SIMD instructions])
+          ;;
+        4)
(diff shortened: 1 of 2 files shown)


mirabilos:
This will not work with x32 (but neither does the old code, so no regression, just pointing out): x32 has 4-byte pointers and `long`s but runs in amd64 long mode.

I guess we’ll address this when adding x32 assembly; for now, I just explicitly pass `--disable-simd` when building for x32 (as we are already doing in Debian).

mirabilos:
No, I now really believe this is really a bad idea altogether.

You can use something like `CC="gcc -m32" configure --host=i386-linux-gnu` already, which is more correct (and then, we can recognise x32 as `x86_64-linux-gnux32` just fine). That’s what the canonical GNU autotools triplets are for (things like “x86” would also be normalised or are outright illegal).

proski:
I believe the detection of features is more robust than the classification of the systems by their canonical names. It should be possible to detect x32 and treat is as unsupported by the accelerated code.

If the code doesn't use any assembly, I can use `CC -gcc -m32` and that's sufficient. If the assembly is used, I have to use `--host`. That's not logical.

mirabilos:
Pavel Roskin dixit:

>I believe the detection of features is more robust than the
>classification of the systems by their canonical names. It should be

Depends, but yes.

>possible to detect x32 and treat is as unsupported by the accelerated
>code.

x32 can be detected with:

    #if defined(__x86_64__) && defined(__ILP32__)
    /* this is x32 */
    #endif

>If the code doesn't use any assembly, I can use `CC -gcc -m32` and
>that's sufficient.

Sufficient, but not correct because then, autoconf thinks you’re
building for 64bit while you’re in fact building for 32bit. The
internal things used by e.g. libtool will be all wrong, so you
always must use --host correctly (“normally”).

bye,
//mirabilos
-- 
<ch> you introduced a merge commit        │<mika> % g rebase -i HEAD^^
<mika> sorry, no idea and rebasing just fscked │<mika> Segmentation
<ch> should have cloned into a clean repo      │  fault (core dumped)
<ch> if I rebase that now, it's really ugh     │<mika:#grml> wuahhhhhh

proski:
I assumed that `--host` enables cross-compilation mode. I checked Autoconf manual, it says that both `--build` and `--host` should be specified and set to different values for cross-compilation mode to be enabled. Specifying `--host` alone should be free of unexpected side effects.

So I'm going to close this PR now. I'll submit smaller fixes from this PR as I keep improving the build system.

PR #74, opened 2017-02-06
Use xnfalloc() and xnfcalloc() for memory allocation

proski:
Those functions make the X server exit if memory allocation fails.


metalefty:
Sounds good.

PR #77, opened 2017-02-12
capture per monitor if they are available

jsorg71:


module/rdpCapture.c
```diff
@@ -756,12 +755,12 @@ rdpCapture1(rdpClientCon *clientCon,
         if (ex != 0)
         {
             rect.x2 += ex;
-            if (rect.x2 > min_width)
+            if (rect.x2 > srect.x2)
             {
-                rect.x1 -= rect.x2 - min_width;
-                rect.x2 = min_width;
+                rect.x1 -= rect.x2 - srect.x2;
```
proski: I feel this part became less clear and needs a comment.

module/rdpCapture.c
```diff
@@ -887,11 +860,13 @@ rdpCapture2(rdpClientCon *clientCon,
...
-    rect.x1 = 0;
-    rect.y1 = 0;
-    rect.x2 = min(dst_width, src_width);
-    rect.y2 = min(dst_height, src_height);
+    rect.x1 = src_left;
+    rect.y1 = src_top;
+    rect.x2 = src_left + src_width;
+    rect.x2 = RDPMIN(dst_width, rect.x2);
+    rect.y2 = src_top + src_height;
+    rect.y2 = RDPMIN(dst_height, rect.y2);
```
proski: Can this part be shared between all rdpCaptureN functions?

module/rdpClientCon.c
```diff
@@ -2170,14 +2170,23 @@ rdpDeferredUpdateCallback(OsTimerPtr timer, CARD32 now, pointer arg)
     int num_rects;
     struct image_data id;
 
+    int index;
```
proski: Not a good variable name.

module/rdpClientCon.c
```diff
@@ -2195,30 +2204,114 @@ rdpDeferredUpdateCallback(OsTimerPtr timer, CARD32 now, pointer arg)
...
-                   clientCon->cap_width, clientCon->cap_height,
-                   clientCon->cap_stride_bytes,
-                   clientCon->rdp_format, clientCon->client_info.capture_code))
-    {
-        LLOGLN(10, ("rdpDeferredUpdateCallback: num_rects %d", num_rects));
-        rdpClientConSendPaintRectShmEx(clientCon->dev, clientCon, &id,
-                                       clientCon->dirtyRegion,
-                                       rects, num_rects);
-        free(rects);
+    if (clientCon->dev->monitorCount < 1)
```
proski: Does the non-multimon need to be handled specially? I would prefer to share as much code as possible. I see quite a lot of similar code.

proski: Here's a design pattern I learned from complex projects. Always code for the hardest case. That gives good input for design. Then port to easier cases. Keep additional logic outside the original code by using short functions and macros.

proski:
Why is it good to capture per monitor? How would it benefit users?

jsorg71:
This is what Windows 7.1 RemoteFX does.  I think you don't want to do the whole desktop because less memory and more responsive with smaller tile sets.  Also, don't break up smaller than a monitor because then you see tearing.

proski:
OK, makes sense.

module/rdpClientCon.c
```diff
@@ -2195,30 +2247,92 @@ rdpDeferredUpdateCallback(OsTimerPtr timer, CARD32 now, pointer arg)
...
+            if (band_index == band_count)
+            {
+                /* gone through all bands, nothing changed */
+                rdpRegionDestroy(clientCon->dirtyRegion);
+                clientCon->dirtyRegion = rdpRegionCreate(NullBox, 0);
+            }
+        }
+        else
+        {
+            /* nothing changed in visable area */
```
proski: visable -> visible

module/rdpClientCon.c
```diff
@@ -2162,13 +2162,65 @@ rdpClientConSendPaintRectShmEx(rdpPtr dev, rdpClientCon *clientCon,
 }
 
 /******************************************************************************/
+static int
+rdpCapRect(rdpClientCon *clientCon, BoxPtr cap_rect, struct image_data *id)
```
proski: Would it be possible to add comments to new functions? The name is too short to be clear.

module/rdpCapture.c
```diff
@@ -632,72 +604,70 @@ rdpCapture0(rdpClientCon *clientCon,
         (*out_rects)[i] = rect;
     }
 
-    if ((src_format == XRDP_a8r8g8b8) && (dst_format == XRDP_a8r8g8b8))
+    src = id->pixels;
+    dst = id->shmem_pixels;
+    dst_format = clientCon->rdp_format;
+    src_stride = id->lineBytes;
+    dst_stride = clientCon->cap_stride_bytes;
```
jsorg71: That depends on the version of xorg you are building against and the system you are building on.  Xorg used to be way behind on there compile requirement.  But you're right, the latest one is using C99 stuff.  I think sprinkling variables allover functions is another issue.

module/rdpClientCon.c
jsorg71: Sure, I was thinking that when I added it.

jsorg71:
I'm going to merge this one as people are waiting on it.
We can do a new PR for other items.

PR #90, opened 2017-04-02
Use stdint.h, constify more pointers, fix clang warning

proski:
clang 3.9.1 produces a warning about casting 255 to char. Fixing that warning lead me to making extensive changes throughout the code.

Xorg uses stdint.h, so it should be OK for xorgxrdp. Take advantage of the types that use the exact bit size. It documents the intent of the code much better.

Add `const` to many more pointers.

Fix an issue in RDPCLAMP macro that made it impossible to cast to another type.


metalefty:
I think this is OK to merge, @jsorg71 what do you think?

metalefty:
Let's merge.

PR #94, opened 2017-06-06
keep and use crc calculations for rfx tiles to save bandwidth when an app redraws aggressively

metalefty:
Resolve conflicts after #90.


metalefty:
@jsorg71 merge?

metalefty:
Jay approved to merge this in gitter, merging.

PR #99, opened 2017-07-11
randr change to prevent xrandr app crashes

jsorg71:


metalefty:
xorgxrdp v0.2.2 is already released in time (June, 2017) but let's merge this and make another release.
Let's release xorgxrdp v0.2.2 and xrdp v0.9.3 at the same time.

Issue #51, opened 2016-12-07
support xrdp-dis


metalefty:
I think neutrinolabs/xrdp#440 should be done in xorgxrdp.

> xrdp-dis does not disconnect the session because "not in an xrdp session" due to file /tmp/.xrdp/xrdp_disconnect_display_ not present.

metalefty:
Done by #101.

PR #101, opened 2017-08-03
Implement disconnection by xrdp-dis

metalefty:
Solves #51, neutrinolabs/xrdp#440.


metalefty:
@jsorg71 what do you think of this?

jsorg71:
I think this add a feature that X11rdp has so it's good.
At some point, we can discuss a more extendable control.

PR #106, opened 2017-10-30
Removes the monitor 0 restriction for xrdpdev

jterry75:
Allows the use of monitor 0 for the xorgxrdp device.
Properly checks for the difference between monitor `"0"` and the `0` return from `atoi` on invalid data by switching to `strtol`.


jsorg71:
+1

metalefty:
+1 Good catch.

Issue #107, opened 2017-11-22
xrdp error on centos7.4


moobyfr:
Testing xrdp on centos7.4 I got the following error
(xrdp from git master, xrorgxrdp from git master)

starting Xorg, I get the following in .xorgxrdp.XX.log:
```
[20171122-21:22:13] [ERROR] X error [BadWindow (invalid Window parameter)](3) opcodes 18/0 resource 0x1400166
[20171122-21:22:13] [ERROR] X error [BadWindow (invalid Window parameter)](3) opcodes 25/0 resource 0x1400166
[20171122-21:22:14] [ERROR] X error [BadWindow (invalid Window parameter)](3) opcodes 18/0 resource 0x14001ab
[20171122-21:22:14] [ERROR] X error [BadWindow (invalid Window parameter)](3) opcodes 25/0 resource 0x14001ab
[20171122-21:22:14] [ERROR] X error [BadWindow (invalid Window parameter)](3) opcodes 18/0 resource 0x14001ab
[20171122-21:22:14] [ERROR] X error [BadWindow (invalid Window parameter)](3) opcodes 25/0 resource 0x14001ab
```

And `gnome-session-daemon` makes an error at same time
`traps: gnome-settings-[12123] trap divide error ip:7f56b1f74979 sp:7ffeb71be370 error:0 in libcolor.so[7f56b1f6d000+d000]`

Some non-common entries in `.xorgxrdp.XX.log:`
```
[  2236.189] rdpRRCrtcGetGamma: 0x55a877f4b080 (nil) (nil) (nil)
[  2238.371] rdpkeybChangeKeyboardControl:
[  2238.371] rdpkeybChangeKeyboardControl: autoRepeat off
[  2238.371] rdpkeybChangeKeyboardControl:
[  2238.371] rdpkeybChangeKeyboardControl: autoRepeat off
[  2238.626] rdpRRCrtcGetGamma: 0x55a877f4b080 0x55a87825acd0 0x55a878261f70 0x55a878261fa0
```

moobyfr:
resolved with #111

PR #111, opened 2018-01-02
Prevent GNOME 3 gnome-settings-daemon crash on RHEL/CentOS 7.4

matt335672:
When using xorgxrdp, a call to XRRGetCtrcGamma() returned a result with a single entry.

This is confusing to some software. For example, gnome-settings-daemon on RHEL/CentOS 7.4 can coredump which takes out the whole session. See:-

https://bugs.centos.org/view.php?id=14054

The coredump is caused by a division of (size-1) where 'size' is the number of entries in the gamma ramp.

This change replaces the single entry gamma ramp with a 256 entry gamma ramp, which seems to be more standard. This is what is implemented in tigervnc-server (see https://github.com/TigerVNC/tigervnc/blob/master/unix/xserver/hw/vnc/xvnc.c ) and also by nouveau on a GTX 470.

### DIFF
--- a/module/rdpRandR.c
+++ b/module/rdpRandR.c
@@ -178,19 +178,6 @@ rdpRRCrtcGetGamma(ScreenPtr pScreen, RRCrtcPtr crtc)
 {
     LLOGLN(0, ("rdpRRCrtcGetGamma: %p %p %p %p", crtc, crtc->gammaRed,
            crtc->gammaBlue, crtc->gammaGreen));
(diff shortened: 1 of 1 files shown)


matt335672:
I'm pretty sure this is the same issue as the trap divide error reported in  https://github.com/neutrinolabs/xorgxrdp/issues/107

moobyfr:
Nice catch, I didn't have time to dig the cause of #107

PR #114, opened 2018-02-14
show correct module version in Xorg log

metalefty:
Ideally, the version string in `rdp.h` should be replaced automatically but the module version is defined separatedly, not combined like `0.2.5`.

```C
#define PACKAGE_VERSION_MAJOR 0
#define PACKAGE_VERSION_MINOR 2
#define PACKAGE_VERSION_PATCHLEVEL 5
```

```
AC_INIT([xorgxrdp], [0.2.5], [xrdp-devel@googlegroups.com])
```


metalefty:
Let's work on automatic replacing later.

PR #115, opened 2018-02-14
Don't define package/module version in multiple files

metalefty:
Now configure.ac is the only place to define version. Pass the version
via Makefile and compiler command line args -Dfoo=bar.


metalefty:
Tested on CentOS 7, FreeBSD 11, macOS High Sierra. Working fine.

PR #123, opened 2018-06-12
change XvExtension use

jsorg71:
Autotools configures xorg-server.h
define XvExtension 1
Meson configures xorg-server.h
define XvExtension
So we just want to test for if XvExtension is defined.
If XVideo is not configured, XvExtension is not defined. You never see
define XvExtension 0
so this should work for all Xorg versions.


metalefty:
Let's merge it and release v0.2.7.

Issue #104, opened 2017-10-06
Invalid memory reads and writes


keithrob91:
Hi,

I have run a static analysis tool on xorgxrdp which has shown up a few issues that I hope you don’t mind me reporting. Most of them relate to multiple invalid memory reads and writes, with most of them related to the same code pattern.

They relate to the use of rdpClientConPreCheck() and the fact that calling this function can free ‘clientCon’ and ‘clientCon->out_s’, and then these are used after the call to rdpClientConPreCheck().

For example:

Within rdpClientConSetOpcode() (rdpClientCon.c) the following occurs:

rdpClientConPreCheck ( dev, clientCon, 6 ); <--- clientCon->out_s and clientCon can both be free’d
out_uint16_le ( clientCon->out_s, 14 ); <--- clientCon and clientCon->out_s both accessed

It looks like most, if not all, of these occur under an error condition which causes rdpClientConDisconnect() to be called. This is the function that actually free’s the memory, and all subsequent accesses to this memory are flagged by the tool.

An example of an invalid memory write is within rdpClientConPreCheck() itself after the error condition has occurred and the following line is executed:

clientCon->count = 0;

Additionally, the tool is also reporting an array overrun in rdpmouseControl() when the function InitPointerDeviceStruct() is called. It is complaining that passing 9 for the numButtons argument with a map array of size 9 overruns the array. I suspect this is because InitButtonClassDeviceStruct(), which is called by InitPointerDeviceStruct(), performs a loop from 1 to numButtons and dereferences the array at map[9].

metalefty:
Thanks for reporting!  I'll investigate it when I have time. If you can provide a patch, I'll look at it.

matt335672:
I've had a look at the excellent concerns @keithrob91 has raised with the static analysis tool.

The basic problem seems that whenever `rdpClientConSend()` or `rdpClientConRecv()` detect a connection problem they immediately call `rdpClientConDisconnect()` which frees the `clientCon`. it isn't immediately obvious when they are being called. As a result, `clientCon` could be used after being freed in several locations.

I've got a patch put together which I need to unit test before I submit it. What's the best way to do submit a patch for an issue? Just submit a pull request as usual?

moobyfr:
yes, pull request is the easier method

matt335672:
This was fixed a long time ago - closing

PR #124, opened 2018-07-20
Fix invalid memory access in rdpmouseControl()

matt335672:
This is one of a couple of pull requests to address the memory issues found in #104 by @keithrob91.

This is the simpler of the two and corresponds to the last paragraph of the issue raised.

The second argument of the call to InitPointerDeviceStruct() is an array based at 1, rather than 0. This can be seen in other projects. See for example vncPointerProc() from TigerVNC's [unix/xserver/hw/vnc/Input.c](https://github.com/TigerVNC/tigervnc/blob/master/unix/xserver/hw/vnc/Input.c). The Xorg code on the other end is in [xserver/dix/devices.c](https://cgit.freedesktop.org/xorg/xserver/tree/dix/devices.c)

### DIFF
--- a/xrdpmouse/rdpMouse.c
+++ b/xrdpmouse/rdpMouse.c
@@ -234,11 +234,13 @@ rdpInputMouse(rdpPtr dev, int msg,
 static int
 rdpmouseControl(DeviceIntPtr device, int what)
 {
-    BYTE map[9];
+#define NBUTTONS 9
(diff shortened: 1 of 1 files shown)


matt335672:
Fair point. I've added a loop and used a macro for the number of buttons like everyone else seems to.

If you're happy with that let me know and I'll squash the commits.

metalefty:
Squash it! Then I'll merge.

matt335672:
Squashed.

PR #133, opened 2018-10-20
Autorepeat

jsorg71:
In some cases, the wrong keyboard device autorepeat was disabled because the way TimerSet works.


metalefty:
What does this fix? Autorepeat supposed to be done in RDP but doing in Xorg?

jsorg71:
Yup, autorepeat is handled in RDP but Xorg has one too.  We want the Xorg autorepeat disabled.  The old code was using the timer incorrectly as sometimes one of the keyboard devices was not getting disabled.  You can check if auto-repeat is enabled with
xset q
In an XRDP session if you run
xset r on
It should not let you turn on autorepeat.
I think Ill update the commit log to include this.

PR #138, opened 2019-02-20
Update .travis.yml

zero77:
This PR builds a deb package using Travis CI, which should then automatically be uploaded to git after.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,6 @@ script:
   - ./configure
   - make
   - sudo checkinstall --install=no -y 
-  - ls
   - wget -c https://github.com/probonopd/uploadtool/raw/master/upload.sh
   - bash upload.sh *.deb
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-sudo: false
+sudo: true
 
 language: c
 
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -33,6 +33,13 @@ script:
   - make CFLAGS="$CFLAGS -O2 -Wall -Wwrite-strings -Werror"
   - make check
   - test -z "$DISTCHECK" || make distcheck
+  - sudo apt-get install git autoconf libtool pkg-config gcc g++ make  libssl-dev libpam0g-dev libjpeg-dev libx11-dev libxfixes-dev libxrandr-dev  flex bison libxml2-dev intltool xsltproc xutils-dev python-libxml2 g++ xutils libfuse-dev libmp3lame-dev nasm libpixman-1-dev xserver-xorg-dev checkinstall -y
+  - ./bootstrap
+  - ./configure
+  - make
+  - sudo checkinstall --install=no -y 
+  - wget -c https://github.com/probonopd/uploadtool/raw/master/upload.sh
```
metalefty: this is not acceptable. content of the script can be changed and we cannot guarantee what will be done during travis build test.

zero77: You make a good point i will add ``upload.sh`` to this repository but, lightly in a different PR.
Because, i dont no how to add it in this one.

zero77: Dun so in https://github.com/neutrinolabs/xorgxrdp/pull/139

matt335672:
I'm having to close #138 and #139, for the simple reason that we've had to move away from travis CI to github actions (see #182)

PR #139, opened 2019-02-27
Needed For #138

zero77:
This PR is needed in https://github.com/neutrinolabs/xorgxrdp/pull/138 for it to work.


matt335672:
I'm having to close #138 and #139, for the simple reason that we've had to move away from travis CI to github actions (see #182)

PR #140, opened 2019-04-08
handle suppress message from xrdp

jsorg71:


metalefty:
Regarding xorgxrdp, no changes since the last release. No releases in April.
Feel free to merge if you think the feature is merge-ready.

PR #143, opened 2019-05-03
Idle timer optimized version

pieska:
this version does not use fixed 1s timers. it calculates the remaining idle time and sets the timer to calculated value. this way only the minimum of timer eexcutions are needed.


metalefty:
Good work.

PR #144, opened 2019-05-04
optimize disconnect-timer to one-shot

pieska:
i changed the disconnect timer from 10s-interval to one-shot. disconnecttimer will now be resetted at reconnect and not in the callback anymore.

DisconnectedTimeLimit can now be 0 which means immediate kill (feature needed by me). KillDisconnected still enables the kill feature, so no change in functionality.

### DIFF
--- a/module/rdpClientCon.c
+++ b/module/rdpClientCon.c
@@ -268,6 +268,20 @@ rdpClientConGotConnection(ScreenPtr pScreen, rdpPtr dev)
     }
 #endif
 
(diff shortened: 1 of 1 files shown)


pieska:
I adapted the PR from 2019 to the current code base, is there anything I can do to get it merged?

matt335672:
Sorry @pieska - like many other projects we've been quite busy here dealing with AI-assisted security reports.

This looks OK to me. I'll merge it now.

PR #157, opened 2020-01-14
support button 8/9

okhowang:
associate with [https://github.com/neutrinolabs/xrdp/pull/1478](https://github.com/neutrinolabs/xrdp/pull/1478)

### DIFF
--- a/xrdpmouse/rdpMouse.c
+++ b/xrdpmouse/rdpMouse.c
@@ -128,7 +128,7 @@ PtrAddEvent(rdpPointer *pointer)
         pointer->old_cursor_y = pointer->cursor_y;
     }
 
-    for (i = 0; i < 7; i++)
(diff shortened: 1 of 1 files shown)


okhowang:
can we merge this PR by the way @metalefty 
vnc do not support mouse button 8/9 as it's more than a byte
but xorg can support it

metalefty:
Nice work!

xrdpmouse/rdpMouse.c
```diff
@@ -226,6 +226,22 @@ rdpInputMouse(rdpPtr dev, int msg,
             pointer->button_mask = pointer->button_mask | 64;
             PtrAddEvent(pointer);
             break;
+        case 115:
```
metalefty: BTW, these (existing) magic numbers should be constified but it's OK for now.

PR #172, opened 2020-09-22
support set monitor info and scale

okhowang:
relate https://github.com/neutrinolabs/xrdp/pull/1692

use client info set parameter for xorg.

many window managers have different scale setting method.
see https://wiki.archlinux.org/index.php/HiDPI .
It seems that cannot adapt for all wm.

and there is no other multi monitor dpi solution except https://askubuntu.com/a/555812
but it will break layout of monitors. because it will change resolution of monitors.

so I don't process scale in this PR.

### DIFF
--- a/module/rdpRandR.c
+++ b/module/rdpRandR.c
@@ -41,12 +41,16 @@ RandR draw calls
 #include <fb.h>
 #include <micmap.h>
 #include <mi.h>
+#include <pixman.h>
+#include <xorg/rrtransform.h>
+#include <xorg/picturestr.h>
 
 #include "rdp.h"
 #include "rdpDraw.h"
 #include "rdpReg.h"
 #include "rdpMisc.h"
 #include "rdpRandR.h"
+#include "rdpClientCon.h"
 
 #if defined(XORGXRDP_GLAMOR)
 #include <glamor.h>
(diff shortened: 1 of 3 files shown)


okhowang:
I found use `xf86SetApi` can set global dpi for X system.
but It didn't work if I call it in `rdpRRSetRdpOutputs`.
can we load client_info earlier for compute dpi in `rdpPreInit`

module/rdpRandR.c
```diff
@@ -364,9 +368,32 @@ rdpRRSetPanning(ScreenPtr pScreen, RRCrtcPtr crtc, BoxPtr totalArea,
...
+        return;
+    if (!RROutputSetPhysicalSize(output, monitor->physical_width, monitor->physical_height))
+        LLOGLN(0, ("%s: RROutputSetPhysicalSize failed", func));
+}
+
+static Rotation
+rdpGetRotation(uint32_t orientation) {
+    switch(orientation) {
+        case 90:
+            return RR_Rotate_0;
```
metalefty, line 384: 90?

okhowang: fixed

Nexarian:
We already call `xf86SetDpi` here: https://github.com/neutrinolabs/xorgxrdp/blob/a4ce59b46867e61ffc7bb630824d764916c78832/xrdpdev/xrdpdev.c#L244

As to the right place to call it to make it work for client info, I'm not yet sure. We should play with it. I've also shameless copied your code here: https://github.com/Nexarian/xorgxrdp/pull/new/dpi_resizing_prototype so I can play with it. It is rebased off the latest devel branch.

@jsorg71 Do you know where we should put `xf86SetDpi` for screen resizing purposes? We should also pay attention to this for LRandR use-cases.

PR #177, opened 2020-12-18
give group permission also to disconnect socket

metalefty:
Missed in #110.


metalefty:
Fix mistake and force pushed.

matt335672 (approved):
Looks OK as far as it goes. There's no configurability here though - you can't specify the group owner of the socket, or whether you want to allow this option at all.

PR #181, opened 2021-02-17
Add "DefaultServerLayout" option

matt335672:
Fixes neutrinolabs/xrdp#1784

Another way to do this is to add a `-layout` switch to `/etc/xrdp/sesman.ini`. I'm less keen on this because of the coupling it introduces between the two projects (i.e. xrdp now needs to know the name of the xorgxrdp server layout). Any thoughts on this?

### DIFF
--- a/xrdpdev/xorg.conf
+++ b/xrdpdev/xorg.conf
@@ -7,6 +7,9 @@ Section "ServerLayout"
 EndSection
 
 Section "ServerFlags"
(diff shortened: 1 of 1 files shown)


metalefty:
+1 to this rather than `-layout` switch.

matt335672:
Thanks for that.

When @aquesnel has had a look at my latest change to the CI PR I'll merge that one, then I'll merge this one to check the CI actually gets triggered.

PR #182, opened 2021-02-17
Move to github actions for CI

matt335672:
Use Github actions rather than Travis for CI.

@aquesnel - you're much more familiar with Github actions than I am. Please feel free to review this.


matt335672:
Link to example run:-

https://github.com/matt335672/xorgxrdp/actions/runs/575399461

Link to Travis CI run:-

https://travis-ci.org/github/neutrinolabs/xorgxrdp/builds/759314531

This PR doesn't seem to have triggered an action. I don't understand why not.

.github/workflows/build.yml
```diff
@@ -0,0 +1,75 @@
...
+      - name: Define feature and arch dependent environment variables
+        # Note: any "variable=value" written to the $GITHUB_ENV file will be
+        # defined as an environment variable for all future steps in this job
+        # See: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#setting-an-environment-variable
+        run: |
+            exec >> $GITHUB_ENV
+            echo "XRDP_CFLAGS=-I$GITHUB_WORKSPACE/xrdp/common"
+            if [ -n "${{ matrix.CFLAGS }}" ]; then
+            echo "CFLAGS=${{ matrix.CFLAGS }}"
+            fi
```
aquesnel, line 65: To me using `exec` is a non-obvious way to set the environment variables.

aquesnel (approved):

aquesnel:
I suspect that the `neutrinolabs/xorgxrdp` repo is checked for github actions definitions on pull requests. Since there is no github actions defined yet, that is why no build was triggered.

.github/workflows/build.yml
matt335672: Fair point.

This is a shell-ism to append all future output from the script to $GITHUB_ENV.

As it happens, after a good night's sleep, I was able to remove this run statement completely and move everything to the env variable above. I think this is a better approach. Do you agree?

matt335672:
I've also fixed a few problems with the output from 'make check':-

1. The line that was in there:-

   ```
   for f in tests/*.log; do echo $f; cat -n $f; done
   ```

   doesn't actually work if there are no log files for the glob to expand. A one-liner to fix this was getting complicated, so I used another script.

2. 'make distcheck' runs 'make check', but doesn't keep the log files. So I've added another step to do this.

Link to improved output is:-

https://github.com/matt335672/xorgxrdp/actions/runs/577993332

@aquesnel - since you're following the PR you've probably been subjected t a barrage of emails while I've got this sorted out. Apologies for that.

.github/workflows/build.yml
```diff
@@ -0,0 +1,72 @@
...
+      - run: git clone --depth 1 --branch=devel https://github.com/neutrinolabs/xrdp.git ${{ github.workspace}}/xrdp
+      - run: ./bootstrap
+      - run: ./configure ${{ matrix.CONF_FLAGS }}
+      - run: make CFLAGS="$CFLAGS -O2 -Wall -Wwrite-strings -Werror"
+      - if: ${{ matrix.CHECK }}
+        run: make check
+      - if: ${{ matrix.DISTCHECK }}
+        run: make distcheck
+      - name: Display test logs (make check only)
+        run: scripts/display_test_logs
```
aquesnel, line 72: I think `if: failed()` or `if: always()` is required to make sure the logs get output when there is a failure in a previous step.

matt335672: Quite right - I verified this on a temp branch

.github/workflows/build.yml
aquesnel: yeah, I think this is easier to understand.

aquesnel:
@matt335672 these changes look good, I think this pr should be merged.

PR #184, opened 2021-02-26
Check client_info version from xrdp

matt335672:
Companion PR to neutrinolabs/xrdp#1813

This won't pass CI at the moment for obvious reasons!

Here's the tail of an example xorgxrdp log for a version mismatch:-

```
. . .
[   289.083] expected xrdp client_info version 20210225, got 20210226
[   289.083] (EE)
Fatal server error:
[   289.083] (EE) Incompatible xrdp version detected  - please recompile(EE)
[   289.083] (EE)
Please consult the The X.Org Foundation support
         at http://wiki.x.org
 for help.
[   289.083] (EE) Please also check the log file at ".xorgxrdp.11.log" for additional information.
[   289.083] (EE)
[   289.083] rdpmouseControl: what 4
[   289.083] rdpkeybControl: what 4
[   289.083] rdpLeaveVT:
[   289.084] (EE) Server terminated with error (1). Closing log file.
```

I've gone for a fatal error as I think this will save time in debugging. Thoughts?

### DIFF
--- a/module/rdpClientCon.c
+++ b/module/rdpClientCon.c
@@ -839,6 +839,14 @@ rdpClientConProcessMsgClientInfo(rdpPtr dev, rdpClientCon *clientCon)
     memcpy(&(clientCon->client_info), s->p - 4, bytes);
     clientCon->client_info.size = bytes;
 
+    if (clientCon->client_info.version != CLIENT_INFO_CURRENT_VERSION)
+    {
+        LLOGLN(0, ("expected xrdp client_info version %d, got %d",
+                   CLIENT_INFO_CURRENT_VERSION,
+                   clientCon->client_info.version));
+        FatalError("Incompatible xrdp version detected  - please recompile");
+    }
+
     LLOGLN(0, ("  got client info bytes %d", bytes));
     LLOGLN(0, ("  jpeg support %d", clientCon->client_info.jpeg));
     i1 = clientCon->client_info.offscreen_support_level;


Nexarian (approved):

module/rdpClientCon.c
```diff
@@ -839,6 +839,14 @@ rdpClientConProcessMsgClientInfo(rdpPtr dev, rdpClientCon *clientCon)
     memcpy(&(clientCon->client_info), s->p - 4, bytes);
     clientCon->client_info.size = bytes;
 
+    if (clientCon->client_info.version != CLIENT_INFO_CURRENT_VERSION)
+    {
+        LLOGLN(0, ("expected xrdp client_info version %d, got %d",
+                   CLIENT_INFO_CURRENT_VERSION,
+                   clientCon->client_info.version));
+        FatalError("Incompatible xrdp version detected  - please recompile");
```
aquesnel, line 847: I'm not familiar with `FatalError`, what exactly does  do?

Can the xorgxrdp version number for the client info be sent back to xrdp as a separate message or a disconnect error code? Having the disconnect error code means that the error can be communicated back to the user attempting to login, instead of being in the xorgxrdp log file which is inaccessible to the user.

matt335672: FatalError is an internal Xorg call. It produces the 'EE' lines in the log file :-

```
[   289.083] (EE) Incompatible xrdp version detected  - please recompile(EE)
[   289.083] (EE)
```

I've added a comment below about your disconnect error code.

matt335672:
As discussed on Gitter, I think the disconnect error code @aquesnel has mentioned is a good idea, when we can get the log window working.

I can raise a separate issue for the Xorg log window (which could include the disconnect code), or I could pull that in to scope of this work. I'm quite torn about priorities here. The SCP stuff I'm looking at is important, but I can see that getting the error reporting improved for Xorg would help a lot for the user community.

I'm happy to take advice from anyone (@metalefty, @Nexarian?) as to the best course of action here.

aquesnel:
Hi @matt335672,

I agree that the SCP work is important, and that we don't want to delay it because of increasing scope on tangential work. I'm not familiar with either the details of the SCP work nor xorgxrdp, so I can't give my opinion on what to do, but here are a few questions that I'd ask myself to help make a decision:

* Is adding the disconnect error code work that will be needed for the SCP work anyways?
* Is adding the disconnect error code straight forward and low effort (<1 day)?

If the answer to either question is yes, then I'd expand the scope of this issue, otherwise I'd create a new issue for tracking the disconnect error code work.

Here's my thoughts on the work needed to add the disconnect error code to the log window: I think the work needed is to add a disconnect message type and then  publish the error code to the existing log window from [xrdp_wm.c](https://github.com/neutrinolabs/xrdp/blob/devel/xrdp/xrdp_wm.c#L2147) via the [server_msg()](https://github.com/neutrinolabs/xrdp/blob/devel/xrdp/xrdp_mm.c#L433) function exposed in the module interface. I'm not familiar with the xorgxrdp code so I'm not sure how much effort it would be to do this work.

If you have a different idea please do share it.

matt335672:
Yes, that's pretty much the kind of lines I was thinking along.

I'm not that familiar with xorgxrdp, but around line 244 of xup.c in `lib_mod_connect()` we send a stream of messages to the X server without waiting for any ack at all. When we return 0 from that function, it's game over as we're then into a black screen and we're expecting the X server to just work.

I think we're into additional issue/PR territory now. I'll raise a separate issue on this, and add the detail above. Maybe we can get this tiny enhancement into devel (which is better than nothing) and get more discussions going around the relative fragility of the Xorg startup sequence.

aquesnel (approved):

Issue #171, opened 2020-09-14
xorgxrdp race to paint window


ejolson2005:
It would appear that 

https://github.com/neutrinolabs/xorgxrdp/commit/d9850d1e0065c58829aad6c9d3525c180dc19cb9

may make worse a race condition related to waiting for more updates on slower computers.  What I'm seeing is menu items randomly appearing blank in the fvwm mouse menus and occasional lines missing from an xterm (for example when top is running).  My understanding is that

- The application paints the background.
- xorgxrdp waits 4ms and starts sending the update.
- The application paints the foreground.
- The foreground is missing from the remote  session.

This problem is particularly noticeable when running xorgxrdp-0.2.14 along with xrdp-0.9.14 on a Rock64 single board computer under AARCH64 Ubuntu Linux with the fvwm window manager.  More than 1/2 the time the mouse menus are missing the text.  Similarly, xterm updates while running top exhibit blank lines about 1/10 of the time.

As this appears to be a timing issue, the exact speed of this single-board computer coupled with the way fvwm writes the mouse menus and xterm updates it's window while running top may be important for tracking down and fixing the bug.

As a reference, the same problem is almost non-existent with xorgxrdp-0.2.13 running on the same system.  It is, however, still possible to slide the mouse through an fvwm menu multiple times up and down until the paint repaint sequences lead to an entry in which the background appears but the foreground text does not.  Rather than 1/2 the time, this may happen more like 1/20 of the time (from non-scientific test related to how many times I need to wiggle the mouse before the screen update race results in missing text).

While not having so much missing text is worth one extra frame of latency in my opinion, there may be a better solution that actually fixes this.  I'll be looking at the patch mentioned above and checking whether, for example, an 8ms delay makes a noticeable improvement.  Any help on fixing the root cause, however, would be appreciated.

ejolson2005:
As a follow up, increasing the wait as
```
#define MIN_MS_TO_WAIT_FOR_MORE_UPDATES 8
```
didn't help much.  There is frequently one line of text missing in the fvwm mouse menus.

However, with
```
#define MIN_MS_TO_WAIT_FOR_MORE_UPDATES 30
```
all the text appears in the mouse menus almost every time.  It is still possible to scan though the menu with the mouse multiple times until to highlighting and rewriting leads to a case where the background appears without the foreground text written on it.

My recommendation is to revert the entire patch as the dual timers do not seem useful until the root cause of why the foreground text sometimes goes missing is solved.

[...]

PR #186, opened 2021-03-07
use damage to track any lost screen changes

jsorg71:
fix for #171

### DIFF
--- a/xrdpdev/xrdpdev.c
+++ b/xrdpdev/xrdpdev.c
@@ -87,7 +87,8 @@ Bool g_use_dri3 = TRUE;
   while (0)
 
 static int g_setup_done = 0;
-static OsTimerPtr g_timer = 0;
+static OsTimerPtr g_randr_timer = 0;
+static OsTimerPtr g_damage_timer = 0;
 
 static char g_xrdp_driver_name[] = XRDP_DRIVER_NAME;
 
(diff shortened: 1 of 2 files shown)


matt335672:
This looks clean to me, but I can't pretend to know enough about an X server to comment on the functionality.

@Nexarian - are there any implications on #183 from this? Do we need to merge this one and then re-test yours do you think?

Nexarian:
First, I think some further documentation as to how xorgxrdp works is in order here, as I'll concede I also don't fully understand everything here. What is `damage` and why is it a bug that we need to enable it at all? Is this going to cause any performance degradation?

I will do some testing this week to double-check that this doesn't interfere with the screen resizing patch. However, based on my knowledge of how this works, I don't think this will be a problem. The fix I created is strictly related to direct interactions with the memory buffer. If, in the memory buffer's case, a bad update from damage detection comes in during a resize, well, that's the same issue as the rdpCapture timer firing with a bad update, and exactly what my fix is designed to prevent in the first place.

In fact, it may actually improve resizing. Sometimes there are graphical glitches on resize that "go away" once you start manipulating the screen. The biggest glitch often has to do with the desktop's background looking weird, though that's more of a problem with Mate versus Gnome 3, and I didn't consider it a big of enough deal to investigate.

Nexarian:
My reply was a bit fuzzy on action items, here's what I would expect to close this out:
1. Create another backlog item that captures this concern: https://github.com/neutrinolabs/xorgxrdp/issues/171#issuecomment-758308211 -- So that we can resolve the original user request. That way, we can track debugging why Damage is necessary without blocking them.

1.5: Modify the PR to link to this issue as a `//TODO:`

2. Add some documentation as to what Damage is and why this patch is necessary and what makes it a bug in the first place. This can either be done with (1) or side-by-side in the code, I care not which.

3. I will test that this does not materially impact my changes that I've made to resize. If it does impact it, I propose that we still go forward with this patch as it addresses an actual user need, whereas resize isn't live in XRDP yet, and then I'll work with you all to figure out how to move forward.

I should be able to get that testing done by the end of this week at the absolute latest, and I think the documentation piece should be straightforward, so we're very close to wrapping this up.

Nexarian:
Tested this with https://github.com/Nexarian/xorgxrdp/tree/resize_damage_enhancement -- It works. I wonder if it will impact performance, however. Still, better to fix the bug.

jsorg71:
Let's wait on this one till we're done with resize.  This one is small and easy to rebase.

Nexarian:
@jsorg71 The thing that's now blocking resize is that we're not sure how to fix the integration with NeutrinoRDP. I've looked at it, and it doesn't look easy. I'm not sure if FreeRDP 1.0.1 on which NeutrinoRDP is based supported the dynamic monitor channel.

If that's the case, we need to update to FreeRDP 2.0 to make it work or have some way to disable the resize functionality elegantly with NeutrinoRDP.

xrdpdev/xrdpdev.c
```diff
@@ -723,7 +765,8 @@ rdpScreenInit(ScreenPtr pScreen, int argc, char **argv)
 
     RegisterBlockAndWakeupHandlers(rdpBlockHandler1, rdpWakeupHandler1, pScreen);
 
-    g_timer = TimerSet(g_timer, 0, 10, rdpDeferredRandR, pScreen);
+    g_randr_timer = TimerSet(g_randr_timer, 0, 10, rdpDeferredRandR, pScreen);
+    g_damage_timer = TimerSet(g_damage_timer, 0, 10, rdpDeferredDamage, pScreen);
```
trishume, line 769: Why the 10ms timer here? Is this repeating or just a hack around something not being initialized yet, which could break if 10ms isn't long enough. Maybe add a comment?

jsorg71: I think they are initialized just after the current call.  The 10 can be anything bigger then 0.  If I remember right, 0 does nothing.  Let me see if there is another way to register these things at the right point.

saddy001:
Fixes https://github.com/neutrinolabs/xrdp/issues/1964

Nexarian:
@jsorg71 I see no reason to not merge this in anymore. I think I understand what's going on. Normally all of the calls to the Xorg backend are wrapped by the xorgxrdp driver. However, in spite of all the drawing primitive APIs being wrapped for our own customized damage detection, some changes are not being reported through that mechanism.

This uses the X damage routine which lets X tell us when things have changed, and sometimes we need that in spite of our custom wrapping.

That's annoying, but it doesn't sound like a performance hit. If you still think we need this, I'm good to go with it. @matt335672  @metalefty

matt335672:
@Nexarian - see recent comments on #171.

After your recent work on this, are you still happy to merge this?

Nexarian:
Yes, go ahead.

Nexarian (approved):

Issue #189, opened 2021-04-11
Built error  ‘struct xrdp_client_info’ has no member named ‘version’


jorpilo:
I am trying to build the devel branch however I am getting the following erros: 

rdpClientCon.c: In function ‘rdpClientConProcessMsgClientInfo’:
rdpClientCon.c:842:31: error: ‘struct xrdp_client_info’ has no member named ‘version’
  842 |     if (clientCon->client_info.version != CLIENT_INFO_CURRENT_VERSION)
      |                               ^
rdpClientCon.c:842:43: error: ‘CLIENT_INFO_CURRENT_VERSION’ undeclared (first use in this function                  )
  842 |     if (clientCon->client_info.version != CLIENT_INFO_CURRENT_VERSION)
      |                                           ^~~~~~~~~~~~~~~~~~~~~~~~~~~
rdpClientCon.c:842:43: note: each undeclared identifier is reported only once for each function it                   appears in
rdpClientCon.c:846:42: error: ‘struct xrdp_client_info’ has no member named ‘version’
  846 |                    clientCon->client_info.version));
      |                                          ^
rdpClientCon.c:56:43: note: in definition of macro ‘LLOGLN’
   56 |     do { if (_level < LOG_LEVEL) { ErrorF _args ; ErrorF("\n"); } } while (0)
      |                                           ^~~~~
make[3]: *** [Makefile:677: rdpClientCon.lo] Error 1
make[3]: Leaving directory '/home/jorpilo/.cache/paru/clone/xorgxrdp-devel-git/src/xorgxrdp/module                  '
make[2]: *** [Makefile:696: all-recursive] Error 1
make[2]: Leaving directory '/home/jorpilo/.cache/paru/clone/xorgxrdp-devel-git/src/xorgxrdp/module                  '
make[1]: *** [Makefile:435: all-recursive] Error 1
make[1]: Leaving directory '/home/jorpilo/.cache/paru/clone/xorgxrdp-devel-git/src/xorgxrdp'
make: *** [Makefile:367: all] Error 2

I am assuming this is related to #184 and commit 7ad1981. @matt335672 Is it missing to modify the rdpClientCon.h with the client info? Latest commit of the .h file is Jun

metalefty:
Do you refer `xrdp_client_info` that has `version` member?

jorpilo:
how? where is it defined? I just downloaded the source code and tried to compile it with make.

metalefty:
[...]

PR #192, opened 2021-04-26
bump version to v0.2.16

metalefty:
@matt335672 I've changed dependency to require xrdp v0.9.16. Do you think is this valid?


metalefty:
To shut up issues such as #189 neutrinolabs/xrdp#1868, xrdp version requirements should be bumped.

PR #197, opened 2021-09-20
trigger full screen refresh after initializing shared memory

### LINKED ISSUE #1964: Session reconnect gives black screen until screen changes
This is xrdp v0.9.16, xorgxrdp v*.16, Linux, Guacamole, Firefox client, with initial-program (single application, not desktop) set.

When reconnecting to a session by reloading the browser tab, while not moving the mouse the screen remains black. Only when the mouse is moved AND some GUI element is hovered that triggers a "screen change", for example an input element that changes its appearance on hover, or the mouse cursor changing to the text input cursor, the screen goes from black to showing the actual program. If no such alement exists, e.g. when some modal dialog is blocking the rest of the UI, the black screen cannot be cured. 

Can you reproduce this with a more simple setup? If not, where else would you report this?

FlintRiverSystems:
This should fix https://github.com/neutrinolabs/xrdp/issues/1964

### DIFF
--- a/module/rdpClientCon.c
+++ b/module/rdpClientCon.c
@@ -1026,6 +1026,8 @@ rdpClientConProcessMsgClientInfo(rdpPtr dev, rdpClientCon *clientCon)
         clientCon->shmemstatus = shmemstatus;
     }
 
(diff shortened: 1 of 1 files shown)


metalefty:
@Nexarian Can you have a look at this?

Nexarian:
Looks good to me!

Nexarian (approved):

PR #203, opened 2021-11-30
Ignore screen size changes which don't change anything

matt335672:
As a result of investigating neutrinolabs/xrdp#1928, it was discovered that that when `mstsc.exe` has a single monitor session full-screen, and this session is moved to another monitor and made full-screen, `mstsc.exe` sends two seemingly identical DISPLAYCONTROL_MONITOR_LAYOUT_PDUs for the new screen size in quick succession.

This behaviour was observed on `mstsc.exe` version 10.0.19041.1266. This doesn't seem to be necessary, but we need to deal with it.

[[MS-RDPEDISP]](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpedisp/d2954508-f487-48bc-8731-39743e0854a9) has nothing to say on this, except that the server should react to valid messages.

In practice, what happens is that the two messages cause xorgxrdp to reallocate the shared memory area twice. Between the first allocation and the second allocation, an update is send to xup with the id of the first shared memory area. If xup does not process this mesage until after the second area has been allocated, xup fails to connect to the shared memory area, and xrdp exits.

See also neutrinolabs/xrdp#2065 which attempts to improve xrdp logging in this area.

This PR deals with the above situation by only reallocating the shared memory area if it differs in size from the previous area. In use cases like that mentioned in neutrinolabs/xrdp#1928, where a single-monitor session is moved from one monitor to a monitor of the same resolution, this reallocation will be unnecessary anyway.

@Nexarian - I'd very much appreciate your comments on this, particularly as to whether this is the right way to deal with this. I'll be happy to answer any questions you might have.

### DIFF
--- a/module/rdpClientCon.c
+++ b/module/rdpClientCon.c
@@ -760,7 +760,7 @@ rdpClientConProcessScreenSizeMsg(rdpPtr dev, rdpClientCon *clientCon,
 
     bytes = clientCon->rdp_width * clientCon->rdp_height *
             clientCon->rdp_Bpp;
-    AllocSharedMemory(clientCon, bytes);
+    rdpClientConAllocateSharedMemory(clientCon, bytes);
     clientCon->shmem_lineBytes = clientCon->rdp_Bpp * clientCon->rdp_width;
 
     if (clientCon->shmRegion != 0)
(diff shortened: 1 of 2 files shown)


matt335672:
After some more testing with RFX, I've found a other places where a similar thing can happen.

I've updated the PR to simply reallocate the shared memory area only if the size has changed.

This change introduces some refactoring along the lines of #200. It doesn't however implement the additional logging of that PR.

Nexarian:
@matt335672 This seems like a win to me, but let's make sure we put https://github.com/neutrinolabs/xrdp/issues/1928 to bed before we merge this.

PR #204, opened 2021-11-30
replace non-posix tool `which` with `command -v`

drwilly:
`which` is a non-posix tool that might not be present in restricted
build-environments. It can however be fully replaced by the posix
shell-builtin `command -v`.

See also:
https://lwn.net/Articles/874049/


metalefty:
Sounds reasonable. Thanks for the contribution!

PR #212, opened 2022-03-21
Update for unification of monitor processing

Nexarian:
Related to https://github.com/neutrinolabs/xrdp/pull/1895. Need to update to reflect the changes to xrdp_client_info.h

### DIFF
--- a/module/rdpClientCon.c
+++ b/module/rdpClientCon.c
@@ -971,14 +971,32 @@ rdpClientConProcessMsgClientInfo(rdpPtr dev, rdpClientCon *clientCon)
     {
         LLOGLN(0, ("  client can not do new(color) cursor"));
     }
+/*
+    TODO: Temporary workaround intended to support two different versions of the xrdp_client_info.h
+    header due to a customer request. This should be removed as soon as convenient, probably before the next
+    release. See https://github.com/neutrinolabs/xorgxrdp/issues/217
+*/
+#if CLIENT_INFO_CURRENT_VERSION == 20210723
     if (clientCon->client_info.monitorCount > 0)
+#else
+    if (clientCon->client_info.display_sizes.monitorCount > 0)
+#endif
     {
         LLOGLN(0, ("  client can do multimon"));
+#if CLIENT_INFO_CURRENT_VERSION == 20210723
         LLOGLN(0, ("  client monitor data, monitorCount=%d", clientCon->client_info.monitorCount));
+#else
+        LLOGLN(0, ("  client monitor data, monitorCount=%d", clientCon->client_info.display_sizes.monitorCount));
+#endif
         clientCon->doMultimon = 1;
         dev->doMultimon = 1;
+#if CLIENT_INFO_CURRENT_VERSION == 20210723
         memcpy(dev->minfo, clientCon->client_info.minfo, sizeof(dev->minfo));
         dev->monitorCount = clientCon->client_info.monitorCount;
+#else
+        memcpy(dev->minfo, clientCon->client_info.display_sizes.minfo, sizeof(dev->minfo));
+        dev->monitorCount = clientCon->client_info.display_sizes.monitorCount;
+#endif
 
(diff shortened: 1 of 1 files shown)


aquesnel (approved):

Nexarian:
@matt335672 @metalefty  I'm not sure why the build checks fail. The probably will because of the dependency on xrdp, but additionally, it looks like something is wrong with the build environment. Embarrassingly, I also can't find the "retry/re-run" option.

Nexarian:
Ah I figured it out. Also needed this: https://github.com/neutrinolabs/xrdp/commit/341a32581cb902c49cb557c956d754c912a31d6e

Nexarian:
Decided to separate out the fix for the checks to keep the commits more atomic: https://github.com/neutrinolabs/xorgxrdp/pull/213

matt335672:
Hi @Nexarian 

I've pulled your other fix in, so if you re-base this it should get through the CI now.

Nexarian:
@matt335672 It now fails as expected!

matt335672:
That's fine - we can merge this at the same time as the other changes in neutrinolabs/xrdp#1895.

module/rdpClientCon.c
```diff
@@ -971,14 +971,14 @@ rdpClientConProcessMsgClientInfo(rdpPtr dev, rdpClientCon *clientCon)
     {
         LLOGLN(0, ("  client can not do new(color) cursor"));
     }
-    if (clientCon->client_info.monitorCount > 0)
+    if (clientCon->client_info.display_sizes.monitorCount > 0)
```
metalefty, line 982: Can you keep compatibility with older versions of xrdp? Probably using some macros and check `CLIENT_INFO_CURRENT_VERSION`.

This is not always necessary but I want that at this time.

metalefty (changes requested):

Nexarian:
@matt335672 @metalefty Take another look, and hopefully this can be merged along with https://github.com/neutrinolabs/xrdp/pull/1895

metalefty (approved):

matt335672:
That'll work, and I'm sort of Okay with it, but rather than having three explicit checks against `CLIENT_INFO_CURRENT_VERSION`, it might be better to add a level of indirection:-

```c
/* See <add suitable reference> */
#if CLIENT_INFO_CURRENT_VERSION > 20210723
#   define CLIENT_INFO_HAS_DISPLAY_SIZES
#endif
```

then :-

```c
#if CLIENT_INFO_HAS_DISPLAY_SIZES
    if (clientCon->client_info.display_sizes.monitorCount > 0)
#else
    if (clientCon->client_info.monitorCount > 0)
#endif
```

etc.

What do people think? Is that overthinking it?

Nexarian:
I believe it is overthinking it, largely because:
1. We do not generally intend to have xorgxrdp be backwards compatible. This is a specific, bespoke, one-off request from @metalefty that I believe will be short term.
2. If we DO intend for xorgxrdp to be backwards compatible with varying versions of XRDP, that is a much bigger conversation, and neither what I've written here nor what you propose would be sufficient to deal with that.

My change is a surgical fix to directly fix this issue, and only that. Doing more risks moving the surgery from "minimally invasive" to "actually invasive."

I also expect/suspect we'd want to remove this fix when the next release happens, and this is here only for the interim.

matt335672:
OK - that's fine. You've convinced me.

Do you want to add a comment somewhere along those lines? The comment will obviously also be deleted when we remove this fix!

Nexarian:
@matt335672 Alright, updated. See the issue I've linked to!

matt335672:
@Nexarian, @metalefty - I've added an additional Wiki page for xorgxrdp here:-

https://github.com/neutrinolabs/xorgxrdp/wiki/NEWS

This serves the same purpose as the [NEWS page for xrdp](https://github.com/neutrinolabs/xrdp/wiki/NEWS). We can collect the merged changes as we go along so it makes the actual release process easier.

PR #213, opened 2022-03-21
Fix build break in GitHub workflow

Nexarian:
Need to add `apt-get update` to refresh the aptitude cache before
installing dependencies.

The same fix was introduced in XRDP here:

https://github.com/neutrinolabs/xrdp/blob/dd968a98b097f8331a1ffbd4991130056b913833/scripts/install_xrdp_build_dependencies_with_apt.sh#L128


Nexarian:
@matt335672 @aquesnel @metalefty Take a look.

PR #216, opened 2022-03-25
populate DRI2InfoRec driverNames

akarl10:
https://github.com/neutrinolabs/xorgxrdp/issues/215

Currently needed for VDPAU with intel GPU
Borrowed from xorg modesetting display driver

### DIFF
--- a/xrdpdev/xrdpdri2.c
+++ b/xrdpdev/xrdpdri2.c
@@ -54,6 +54,10 @@ dri2
 #include "rdpPri.h"
 #include "rdpDraw.h"
 
+#if defined(XORGXRDP_GLAMOR)
+#include <glamor.h>
+#endif
+
 #define LLOG_LEVEL 1
 #define LLOGLN(_level, _args) \
   do \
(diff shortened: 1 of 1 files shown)


Nexarian:
@jsorg71 What do you think?

jsorg71:
The coding style is a bit off but I think because modesetting does this, it right.

jsorg71:
I see the latest also includes crocus, although I'm not sure what that is.

```
        if (strcmp(driver_names[0], "i965") == 0 ||
            strcmp(driver_names[0], "iris") == 0 ||
            strcmp(driver_names[0], "crocus") == 0) {
            driver_names[1] = "va_gl";
        } else {
            driver_names[1] = driver_names[0];
        }
```

akarl10:
> I see the latest also includes crocus, although I'm not sure what that is.
> 
> ```
>         if (strcmp(driver_names[0], "i965") == 0 ||
>             strcmp(driver_names[0], "iris") == 0 ||
>             strcmp(driver_names[0], "crocus") == 0) {
>             driver_names[1] = "va_gl";
>         } else {
>             driver_names[1] = driver_names[0];
>         }
> ```

You are right. Seems my source was not on the latest version.
https://gitlab.freedesktop.org/xorg/xserver/-/commit/a7b0a7fabd137183cc42a5edb15697e354c4450c

I will include this change.

About the coding style: it is, as mentioned in the comment, a copy of modesetting's dri.c

Nexarian:
@matt335672 @metalefty I'm OK with this change as it seems Jay is. As we integrate GFX it will definitely need to be updated, but this is a win for now.

matt335672:
@paravz @akarl10 - see also #219 which should prevent this in the future.

akarl10:
> @paravz @akarl10 - see also #219 which should prevent this in the future.

👍 Makes definitely sense

PR #224, opened 2022-07-09
Support set monitor info and scale

Nexarian:
This is a shameless copy of
https://github.com/neutrinolabs/xorgxrdp/pull/172 for my own prototyping
and editing purposes.

We need to figure out how to get xf86SetDpi working.

### DIFF
--- a/module/rdpRandR.c
+++ b/module/rdpRandR.c
@@ -41,12 +41,16 @@ RandR draw calls
 #include <fb.h>
 #include <micmap.h>
 #include <mi.h>
+#include <pixman.h>
+#include <xorg/rrtransform.h>
+#include <xorg/picturestr.h>
 
 #include "rdp.h"
 #include "rdpDraw.h"
 #include "rdpReg.h"
 #include "rdpMisc.h"
 #include "rdpRandR.h"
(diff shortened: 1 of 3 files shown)


module/rdpRandR.c
```diff
@@ -532,29 +577,40 @@ rdpRRSetRdpOutputs(rdpPtr dev)
...
+        struct monitor_info monitor;
+        memset(&monitor, 0, sizeof(monitor));
+        monitor.top = 0;
+        monitor.left = 0;
+        monitor.right = dev->width - 1;
+        monitor.bottom = dev->height - 1;
+        monitor.physical_width
+            = clientCon->client_info.display_sizes.session_width;
+        monitor.physical_height
+            = clientCon->client_info.display_sizes.session_height;
```
matt335672, line 589: I can see where your comment about putting the physical size into neutrinolabs/xrdp#2310 is coming from now.

As I said in response to your comment, that only makes sense in a single monitor scenario. Let me know what your thoughts are in this area.

matt335672 (commented):
That all looks fine to me, when between us we've figured out what to do with the physical size from the core message.

PR #227, opened 2022-08-31
mouse: workaround for too fast vertical scroll

metalefty:
Provided by @seflerZ at neutrinolabs/xorgxrdp#150.

I tested the patch, 10ms interval looks enough.

----

To enable this hack, edit sesman.ini like this:

```
[SessionVariables]
(some existing lines)
XRDP_XORG_TOUCHPAD_SCROLL_HACK=yes
```


matt335672 (approved):
That looks fine to me.

I'm happy to go with 10ms. If it turns out to need tweaking later, we can get the environment variable to take a numeric argument as an option, but at the moment there's no need to over-complicate it.

PR #228, opened 2022-09-02
Simple maintenance improvements

a1346054:


Nexarian:
@a1346054 First time contributor! Welcome! Tell us more about how you use XRDP and what you want to see from it. We have plenty of projects you can help with.

a1346054:
I use xrdp to have access to my linux machine remotely from windows machines without having to install any additional software on the windows machine.

PR #231, opened 2022-09-06
Bump version to v0.9.19

metalefty:


matt335672:
LGTM - the bump for v0.2.18 is d81fe5f342b218cd8efb28007b14a73cc1bdc094, and this looks identical

metalefty:
I created some cherry-pick PRs from devel to v0.9 branch. #232 is the last. 
Any other thoughts?

matt335672:
I can't see anything else in devel which is worth taking over. The only other significant change from devel is 600095a2d5133a13409395b6629309506963672a, and you don't need that.

metalefty:
That's right. Merging.

PR #232, opened 2022-09-06
[v0.9] Ignore screen size changes which don't change anything (#203)

metalefty:
* Only reallocate shared memory if the size changes
Co-authored-by: Nexarian <cmp@pitstick.net>


matt335672 (approved):
That's fine - I tested this when looking at neutrinolabs/xrdp#1928

Issue #235, opened 2022-09-18
Compiling errors due to undeclared constants WM_TOUCH_VSCROLL/WM_TOUCH_HSCROLL


saulvin:
trying to compile xorgxrdp on ArchLinux to support enhanced session mode in an Hyper-V environment results in the following compiler errors.

CC       rdpMouse.lo  
rdpMouse.c: In function 'rdpInputMouse':  
rdpMouse.c:282:14: error: 'WM_TOUCH_VSCROLL' undeclared (first use in this function)  
  282 |         case WM_TOUCH_VSCROLL:  
      |              ^~~~~~~~~~~~~~~~  
rdpMouse.c:282:14: note: each undeclared identifier is reported only once for each function it appears in  
rdpMouse.c:285:14: error: 'WM_TOUCH_HSCROLL' undeclared (first use in this function)  
  285 |         case WM_TOUCH_HSCROLL:  
      |              ^~~~~~~~~~~~~~~~  
make[2]: *** [Makefile:466: rdpMouse.lo] Error 1  

probably the following dependency needs a version bump
xrdp to 0.9.20 as this seems to define those required new constants

metalefty:
Which xrdp version do you want to use?

metalefty:
I already mentioned in the release note. If you want to use xrdp 0.9.20, use xorgxrdp 0.9.x.
https://github.com/neutrinolabs/xorgxrdp/releases/tag/v0.9.19

Your compile error occurs because you're compiling `devel` branch nevertheless you're using xrdp 0.9.x.

| xrdp | xorgxrdp | compile | 
|--|--|--|
|0.9.20|0.9.19|✅|
|devel|devel|✅|
|0.9.20|devel|⛔ <-- your case appears this|
|devel|0.9.19|⛔|

(Current devel branches have the tentative version 0.9.80)

saulvin:
@metalefty yes according to your table that makes sense.

However looking at [configure.ac](https://github.com/neutrinolabs/xorgxrdp/blob/devel/configure.ac) in devel branch of xorgxrdp
versions starting with 0.9.16 seems to satisfy the version dependency to xrdp.

ArchLinux uses is 09.19-2 (according to their package versioning) which also would satisfy this condition.

metalefty:
I have just raised #236 to declare version dependency correctly.

PR #236, opened 2022-09-19
Bump required xrdp version to devel version

metalefty:
Address #235.


matt335672 (approved):
LGTM - sorry I missed this earlier.

Issue #237, opened 2022-09-22
glamor_egl_get_driver_name not defined in Debian 10


laigor:
xorgxrdp v0.9.19 not started in Debian 10 because glamor_egl_get_driver_name not defined.
`Failed to load /usr/lib/xorg/modules/drivers/xrdpdev_drv.so: /usr/lib/xorg/modules/drivers/xrdpdev_drv.so: undefined symbol: glamor_egl_get_driver_name`
I looked at the file /usr/include/xorg/glamor.h and made sure that glamor_egl_get_driver_name is not present in Debian 10 and present in Debian 11.

matt335672:
`glamor_egl_get_driver_name` is defined in `/usr/lib/xorg/modules/libglamoregl.so`:-

```
$ nm -D /usr/lib/xorg/modules/libglamoregl.so | grep glamor_egl_get_driver_name
00000000000092b0 T glamor_egl_get_driver_name
```

On Debian 10, this symbol is not defined, hence the error.

On the face of it, it looks like you're trying to run `xorgxrdp` compiled on Debian 11 on Debian 10. This isn't a supported operation.

Or am I missing something?

laigor:
On debian 10 command
```
$ nm -D /usr/lib/xorg/modules/libglamoregl.so | grep glamor_egl_get_driver_name
$
```
gives empty result.
xorgxrdp compiled on Debian 10

laigor:
Debian 11:
```
$ grep glamor_egl_get_driver_name /usr/include/xorg/glamor.h
extern _X_EXPORT const char *glamor_egl_get_driver_name(ScreenPtr screen);
```
Debian 10:
```
$ grep glamor_egl_get_driver_name /usr/include/xorg/glamor.h 
$
```

matt335672:
That's right, and is more-or-less what I've just said.

If you want to run xrdp 0.9.19 on Debian 10, you'll need to build it on Debian 10 without the `--enable-glamor` flag. It looks like your executable has been built against a later version of the X server which has the necessary symbols defined.

Does that make sense?

laigor:
#238

matt335672:
@laigor - is this now fixed with #239?

PR #238, opened 2022-09-26
Checking if glamor_egl_get_driver_name exists

laigor:
#237

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -28,6 +28,7 @@ if test "x${enable_glamor}" = "xyes"; then
   PKG_CHECK_MODULES([XORG_SERVER_GBM], [gbm >= 0], [], [AC_MSG_ERROR([please install libgbm-dev or mesa-libgbm-devel])])
   PKG_CHECK_MODULES([XORG_SERVER_EPOXY], [epoxy >= 0], [], [AC_MSG_ERROR([please install libepoxy-dev or libepoxy-devel])])
   PKG_CHECK_MODULES([XORG_SERVER_EGL], [egl >= 0], [], [AC_MSG_ERROR([please install libegl1-mesa-dev or mesa-libEGL-devel])])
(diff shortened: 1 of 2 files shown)


jsorg71:
Thanks, We do need a check for this.
This looks like it was added in 1.20.7 so lets use
#if XORG_VERSION_CURRENT >= XORG_VERSION_NUMERIC(1, 20, 7, 0, 0)
instead so we don't need to use PKG_CHECK_EXISTS(epoxy >= 1.5.4...

matt335672:
@laigor - are you able to make the change @jsorg71 mentioned to your PR?

Thanks.

jsorg71:
I can imagine a system with epoxy 1.5.4 and xorg-server 1.20.6 and this PR will still fail.
Lets check xorg-server version.
I created PR #239 for this.
Thanks for pointing this out.

laigor:
> Thanks, We do need a check for this. This looks like it was added in 1.20.7 so lets use #if XORG_VERSION_CURRENT >= XORG_VERSION_NUMERIC(1, 20, 7, 0, 0) instead so we don't need to use PKG_CHECK_EXISTS(epoxy >= 1.5.4...

This is how the xorg xserver checks for the presence of the glamor_egl_get_driver_name function:
```
conf_data.set('GLAMOR_HAS_EGL_QUERY_DRIVER',
              epoxy_dep.found() and epoxy_dep.version().version_compare('>= 1.5.4') ? '1' : false)

```
[Please see this link](https://gitlab.freedesktop.org/xorg/xserver/-/blob/master/include/meson.build)

jsorg71:
````
> conf_data.set('GLAMOR_HAS_EGL_QUERY_DRIVER',
>               epoxy_dep.found() and epoxy_dep.version().version_compare('>= 1.5.4') ? '1' : false)
````

xorg-server version 1.20.7+ does that to check to see if function eglGetDisplayDriverName() exists, not glamor_egl_get_driver_name().
The function we call glamor_egl_get_driver_name() always exists in xorg-server version 1.20.7+.  The check controls whether glamor_egl_get_driver_name() returns NULL or calls eglGetDisplayDriverName().

see
https://gitlab.freedesktop.org/xorg/xserver/-/blob/master/glamor/glamor_egl.c#L707

jsorg71:
fixed  via #239

PR #239, opened 2022-10-12
check for xorg version if glamor_egl_get_driver_name exists

jsorg71:


matt335672:
LGTM. The call was introduced [here](https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/351) [gitlab.freedesktop.org] and this was merged for 1.20.7 of the xorg X server.

PR #241, opened 2022-11-04
Disable some auto-add hardware features

matt335672:
Addresses neutrinolabs/xrdp#2379 and neutrinolabs/xrdp#2253.

On some systems, GPUs are auto-loaded. These can be incompatible with the xorgxrdp module.

These keywords are not used by us yet, so I've done a quick analysis of contemporary systems. The reason is that these can cause errors on systems which don't expect them.

The `AutoAddGPU` keyword was introduced for xorg 1.13.0. This is no longer used by any contemporary systems.

The GPUDevice keyword was introduced for xorg 1.18.0. This is still used by Ubuntu 16.04 which is technically still supported if you have an ESM contract with Canonical.

Our other long-term enterprise system, RHEL 7, uses Xorg 1.20.

I'm marking this draft for now, as I want to test on Ubuntu 16 and CentOS 7.

@akarl10 - do you have any comments on this PR? I'm pretty sure it's what we've discussed recently.

### DIFF
--- a/xrdpdev/xorg.conf
+++ b/xrdpdev/xorg.conf
@@ -12,6 +12,7 @@ Section "ServerFlags"
     Option "DefaultServerLayout" "X11 Server"
     Option "DontVTSwitch" "on"
     Option "AutoAddDevices" "off"
+    Option "AutoAddGPU" "off"
 EndSection
 
 Section "Module"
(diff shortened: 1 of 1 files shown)


akarl10:
Putting the comment about < Xorg 1.18.0 is a good idea 👍
I think being configuration compatible with ancient systems is good, but the user who compiles xrdp on such a old system is probably able to read the comment in xorg.conf. If someone makes packages for this old release it imho it is the responsability of that one to deliver a valid configuration

matt335672:
I've tested this with Ubuntu 16.04 LTS and CentOS 7 now, so I've removed the draft status.

Have I missed anything?

metalefty:
+1

PR #254, opened 2023-03-08
Create a RandR output before the first client connects

matt335672:
Prompted by neutrinolabs/xrdp#2580

This allows clients expecting RandR outputs to work correctly if no client is connected at startup (i.e. if `xrdp-sesrun` is used)

### DIFF
--- a/xrdpdev/xrdpdev.c
+++ b/xrdpdev/xrdpdev.c
@@ -491,6 +491,7 @@ rdpDeferredRandR(OsTimerPtr timer, CARD32 now, pointer arg)
     }
 
     RRScreenSetSizeRange(pScreen, 256, 256, 16 * 1024, 16 * 1024);
+    rdpRRSetRdpOutputs(dev);
     RRTellChanged(pScreen);
 
     return 0;


metalefty (approved):

Nexarian (approved):

Nexarian:
This feels reasonable, and may have good side effects for GFX as well...

Nexarian:
@jsorg71 What do you think? Will this have any side effects for the Nvidia integration or LRandr?

jsorg71:
I'll have to check and make sure there is always a default output when LRANDR is used.
I do have a few comments about waitforx I'll make over there.

jsorg71:
No code in xrdpdev is used when nvidia wrapper is used.  We can make sure LRANDR has a default output later.

jsorg71 (approved):
Looks good

matt335672:
@jsorg71 - as regards waitforx, I'm hoping to make use of this to validate the X server before starting the session.

The intention is to remove a whole class of "blue screen" problems where the X server doesn't start properly, and xrdp then spends 30 seconds trying to connect to it. At the moment it takes specialist knowledge to distinguish this situation from a similar one where the session fails to start.

Details and screenshots at neutrinolabs/xrdp#2592. This is still in draft, but should be ready for review in the next few days.

metalefty:
@matt335672 Should we backport this to v0.9?

matt335672:
I don't think so. No-one's complained about it, and you can't use `xrdp-sesrun` very easily in that version to start unattended sessions.

metalefty:
Thanks, then inertial scrolling is the only change included in the next xorgxrdp v0.9 release. I'll go forward.

PR #256, opened 2023-04-21
work on large cursor

jsorg71:


module/rdpCursor.c
```diff
@@ -228,54 +228,110 @@ rdpSpriteSetCursorCon(rdpClientCon *clientCon,
...
+    }
+    cur_mask = cur_data + 96 * 96 * 4;
+    client_max_width = 32;
+    client_max_height = 32;
+    sending_bpp = 0;
+    can_do_new = clientCon->client_info.pointer_flags & 1;
+    can_do_large = 0;
+#if defined(CLIENT_INFO_CURRENT_VERSION)
+#if CLIENT_INFO_CURRENT_VERSION >= 20230423
+    can_do_large = clientCon->client_info.large_pointer_support_flags & 1;
```
matt335672: If you tested `clientCon->cursor_shmemptr` here, you could maybe fall back to 32x32 if the `shmat()` had failed for some reason. This might be preferable to sending no cursor at all.

matt335672 (commented):
Again, one minor comment.

module/rdpCursor.c
jsorg71: Good point, I'll add that check.  Does it looks like I used CLIENT_INFO_CURRENT_VERSION right?  It's the first time I used that guy.

jsorg71: ok, done

matt335672: I'd be tempted to leave out the `#if defined(CLIENT_INFO_CURRENT_VERSION)` stanza. Other bits of rdpClientCon.c assume it's defined, particularly the version check.

Alternatively, we could add this at the start of the file if you want to run this version of xorgxrdp with xrdp versions before v0.9.16:-

```c
#ifndef CLIENT_INFO_CURRENT_VERSION
#define CLIENT_INFO_CURRENT_VERSION 0
#endif
```

We'd need to make other changes too, however, so it could turn into a bit of a rabbit hole.

module/rdpCursor.c
```diff
@@ -280,13 +280,12 @@ rdpSpriteSetCursorCon(rdpClientCon *clientCon,
     client_max_height = 32;
     sending_bpp = 0;
     can_do_new = clientCon->client_info.pointer_flags & 1;
-    can_do_large = 0;
-#if defined(CLIENT_INFO_CURRENT_VERSION) && defined(LARGE_CURSOR_FLAG_96x96)
-#if CLIENT_INFO_CURRENT_VERSION >= 20230423
+#if (CLIENT_INFO_CURRENT_VERSION >= 20230423) && defined(LARGE_CURSOR_FLAG_96x96)
```
matt335672: Sorry - should have mentioned. If CLIENT_INFO_CURRENT_VERSION >= 20230423, you probably don't need to bother with the `defined(LARGE_CURSOR_FLAG_96x96)` check as that's implicit.

jsorg71: Actually, I added LARGE_CURSOR_FLAG_96x96 a couple of days later.  I'll update that.  So we use CLIENT_INFO_CURRENT_VERSION for all the installed headers, not just struct xrdp_client_info.  Good to know.  When I change it, I'll add a comment.

PR #257, opened 2023-05-03
Remove unused funcs and member variables

matt335672:
In the process of looking at xorgxrdp on an unrelated issue, I found a few unused funcs and member variables.

Can these be removed?


Nexarian:
@jsorg71 This looks good to me, but I want your feedback on it.

jsorg71:
Yes I think we should rebase and merge this.
I'll do a PR for my posix shm changes after that.

matt335672:
Rebased and remerged. Thanks both.

PR #258, opened 2023-05-13
add large cursor support, posix shm

jsorg71:


matt335672 (commented):
See neutrinolabs/xrdp#2670

matt335672 (approved):

PR #261, opened 2023-05-30
move to posix shm

jsorg71:


matt335672:
Excellent - thanks for starting on this one, @jsorg71.

module/rdpClientCon.c
```diff
@@ -687,25 +687,38 @@ rdpClientConProcessMsgVersion(rdpPtr dev, rdpClientCon *clientCon,
...
         return;
     }
-
-    if (clientCon->shmemptr != 0)
+    if (clientCon->shmemfd != -1)
     {
-        shmdt(clientCon->shmemptr);
+        g_free_unmap_fd(clientCon->shmemptr,
+                        clientCon->shmemfd,
+                        clientCon->shmem_bytes);
```
matt335672, line 706: Suggest adding the following in case the allocation later fails:-
```c
        clientCon->shmemptr = NULL;
        clientCon->shmemfd = -1;
        clientCon->shmem_bytes = 0;
```

matt335672 (commented):
Just the one comment from me:-

module/rdpClientCon.c
jsorg71: yes, I think this suggestion also relieved a leak.

PR #267, opened 2023-09-25
Support OpenIndiana

mstone2001:
For this I just added the build flag _XOPEN_SOURCE=600 to the Makefiles.  Found this was the simplest and preferred way to get the new POSIX calls to build on OpenIndiana.  Should also work on Solaris, but don't have any Solaris boxes to test this on.


matt335672:
Thanks for this @mstone2001 

Looks good to me. can you squash and I'll merge.

mstone2001:
Changes are squashed.

PR #273, opened 2024-01-08
Initial work for GFX and mainline merge integration.

Nexarian:


metalefty:
#282 fixed my issue with GFX multimon.

metalefty:
Let's merge into devel.

PR #274, opened 2024-01-18
configure logging and timer changes

jsorg71:


jsorg71:
@Nexarian can you take the first 2 commit from this PR?  I think you will agree we need both.

Nexarian:
I'll just merge everything; clearly my KEY_FRAME fix didn't completely address everything, so let's give this a shot!

PR #276, opened 2024-01-21
fix for green boxes with GFX and glamor

jsorg71:


Nexarian:
For a second I thought "wait didn't we already merge this?" Then I realized it was for EGL!

Nexarian (approved):

PR #278, opened 2024-01-28
GFX: work on multimon

jsorg71:


Nexarian:
There's an unresolved conflict.

Also, thinking about this, most of the commits have pretty generic names, would it be possible to give them descriptions that are a bit more important and/or consolidate them as to their function? Presumably we'll do a direct lift-and-shift of `gfx_mainline_merge_work` into `devel`  we'll want the commits to be meaningful later.

Nexarian:
Also, could you rebase? I've added a few more commits.

metalefty:
Replaced by #278, #279.

PR #284, opened 2024-02-08
Don't delete RandR outputs and CRT controllers

matt335672:
Rebase of #283 

This change doesn't delete outputs or CRT controllers when the screen configuration changes. Instead, the outputs are disconnected.

This emulates the behaviour of other virtual X servers (e.e. tigervnc) and prevents X applications from encountering BadRROutput or BadRRCtrc errors.

The following command can now be used to monitor RandR events generated by xorgxrdp. Previously it could fail with BadRROutput when reconnecting from a client with fewer screens.

```
xev -root -event randr [-display :xx]
```

### DIFF
--- a/module/rdpRandR.c
+++ b/module/rdpRandR.c
@@ -365,38 +365,15 @@ rdpRRSetPanning(ScreenPtr pScreen, RRCrtcPtr crtc, BoxPtr totalArea,
 }
 
 /******************************************************************************/
-static RROutputPtr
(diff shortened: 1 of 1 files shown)


jsorg71:
tested with 2 monitors, looks good to me

matt335672:
Thanks.

I'll probably pull it into the resize rework I'm doing at the moment then.

PR #292, opened 2024-03-11
Bump version to v0.9.20

metalefty:
This version requires xrdp v0.9.25 which supports inertial scrolling.

ref. https://github.com/neutrinolabs/xrdp/pull/2948


matt335672:
Looks much the same as #231 to me - I'm happy with this.

Issue #294, opened 2024-03-12
0.10 version AUR not starting


pwn0r:
Since AUR version of both beta xrdp and xorgxrdp was released on AUR, decided to try. 

Previously xrdp+xorg was working fine on that machine, using recommended modifications (xinitrc and xwrapper). 

Now even uninstalling xrdp and xorgxrpd and installing beta versions the x server does not start. 
Also used a hotfix for waitforx as in one of the issues on xrdp repo. 

I guess the xrdp package is currently broken on AUR? 

The main problem is always the same, the x server is not starting. Trying various things but the result is always same. 
e.g. used both Xorg and /usr/lib/Xorg in the sesman.ini etc.  Does not matter if you remote login right after VM restart or after logging on console first (which works). 

```
[   275.571] 
X.Org X Server 1.21.1.11
X Protocol Version 11, Revision 0
[   275.571] Current Operating System: Linux 6.7.7-1-MANJARO #1 SMP PREEMPT_DYNAMIC Fri Mar  1 18:26:06 UTC 2024 x86_64
[   275.571] Kernel command line: BOOT_IMAGE=/vmlinuz-6.7-x86_64 root=UUID=8a2888af-bced-4bae-aa7b-da0b9159a080 rw quiet udev.log_priority=3
[   275.571]  
[   275.571] Current version of pixman: 0.43.4
[   275.571] 	Before reporting problems, check http://wiki.x.org
	to make sure that you have the latest version.
[   275.571] Markers: (--) probed, (**) from config file, (==) default setting,
	(++) from command line, (!!) notice, (II) informational,
	(WW) warning, (EE) error, (NI) not implemented, (??) unknown.
[   275.571] (++) Log file: ".xorgxrdp.10.log", Time: Tue Mar 12 11:24:42 2024
[   275.571] (++) Using config file: "/etc/X11/xrdp/xorg.conf"
[   275.571] (==) Using config directory: "/etc/X11/xorg.conf.d"
[   275.571] (==) Using system config directory "/usr/share/X11/xorg.conf.d"
[   275.571] (**) Option "defaultserverlayout" "X11 Server"
[   275.571] (**) ServerLayout "X11 Server"
[   275.571] (**) |-->Screen "Screen (xrdpdev)" (0)
[   275.571] (**) |   |-->Monitor "Monitor"
[   275.571] (**) |   |-->Device "Video Card (xrdpdev)"
[   275.571] (**) |-->Input Device "xrdpMouse"
[...]

PR #295, opened 2024-03-13
Improve --enable-glamor behaviour

matt335672:
See #294 

This PR makes the following improvements to the xorgxrdp build process when `--enable-glamor` is used:-

1) Adds `$LIBDRM_CFLAGS` to the include path (where required) to ensure drm_fourcc.h can be located
2) Adds an explicit link dependency on the Xorg module libglamoregl.so to libxorgxrdp.so.
3) Adds the Xorg module path to the RPATH for libxorgxrdp.so, so the run-time linker can find libglamoregl.so
4) 'make check' is now added to the CI for --enable-glamor builds. On Linux this checks that all module symbols can be resolved.

Essentially this patch upstreams [this patch file](https://aur.archlinux.org/cgit/aur.git/tree/glamor.patch?h=xorgxrdp-glamor&id=347fe5c22895e510e32ea2136a896fd14f080d59) from Arch AUR which was created (I believe) by @jat001

@jat001, @Chocobo1 - please feel free to comment


jsorg71 (approved):
LGTM

Issue #300, opened 2024-04-01
xorgxrdp uses high CPU in CRC


metalefty:
GFX (v0.10) uses CPU significantly higher compared to v0.9.

Sample web page to reproduce:

* https://people.freebsd.org/~meta/tmp/whitebox.html

The page draws random rectangles to a canvas, then does the same after clicking "Press to toggle" but draws white on white so it is invisible screen updates. The hotspot is `crc_process_data`.

I'm still investigating why GFX does more CRC calculation and if we can reduce CRC calculation or accelerate using a faster algorithm or library such as [crc32c](https://github.com/google/crc32c).

@jsorg71 @Nexarian Any thoughts?

metalefty:
I found a benchmark comparing CRC algorighm.
* https://github.com/htot/crc32c?tab=readme-ov-file#benchmarks

Nexarian:
Yup this has been a known issue with RFX for a while. The CRC hashing algorithm is slow. One way to address this has been https://github.com/neutrinolabs/xorgxrdp/pull/167, (using other hashing algorithms) which I incorporated into `mainline_merge` but left out because it wasn't fully tested and merged originally.

We need to investigate other hashing codecs that will be better, and I think there is also room to use SIMD optimized hashes for even more performance.

Nexarian:
If you'd like I can get the relevant bits from https://github.com/neutrinolabs/xorgxrdp/pull/167 rebased again off of the latest and we can see if it helps?

jsorg71:
When I run that page with Xorg glamor, it's using almost no CPU, hehe.
I don't really see why GFX should be more than RFX.  I'll test that.  They should both be using the slow CRC32.  We can switch to something else like a hash.  Don't worry about the EGL / glamor CRC stuff.  That will all get moved to the helper later.

jsorg71:
I got working  SSE2 accel for librfxcodec progressive.  I still have to accel the RGB to YUV in Xorg.  I think the CRC change to hash and SSE improvements should really help.

metalefty:
Thank you both. I remembered #167 now.

 I thought CRC is required by protocol specification but it sounds no. We can try whatever fast hash if there is no bound to use CRC. I think we should use libraries which has built-in SIMD accelerations. I'll try some hash algorithms including wyhash.

metalefty:
> When I run that page with Xorg glamor, it's using almost no CPU, hehe.

I see CRC checking is skipped using glamor.
[...]

PR #301, opened 2024-04-02
Switch to wyhash from CRC for capture tile diff 

metalefty:
This is a cut out of #167 and a effort to address #300, originally suggested by @trishume. Removed scroll detection for simplicity at the moment.

After switching to wyhash, CPU usage is reduced by 1/2 to 1/3. I think there is more room to reduce using other hash algorithm with hardware acceleration.

### DIFF
--- a/module/rdpCapture.c
+++ b/module/rdpCapture.c
@@ -591,6 +591,23 @@ isShmStatusActive(enum shared_memory_status status) {
 }
 
 /******************************************************************************/
+/* copy rects with no error checking */
+static uint64_t
+wyhash_rfx_tile(const uint8_t *src, int src_stride, int x, int y, uint64_t seed)
+{
+    int row;
+    uint64_t hash;
+    const uint8_t *s8;
+    hash = seed;
+    for(row = 0; row < 64; row++)
+    {
+        s8 = src + (y+row) * src_stride + x * 4;
+        hash = wyhash((const void*)s8, 64 * 4, hash, _wyp);
+    }
+    return hash;
+}
+
+/******************************************************************************/
 static Bool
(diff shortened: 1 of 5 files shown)


module/rdpCapture.c
```diff
@@ -586,6 +590,23 @@ isShmStatusActive(enum shared_memory_status status) {
...
+    int row;
+    uint64_t hash;
+    const uint8_t *s8;
+    hash = seed;
+    for(row = 0; row < 64; row++)
+    {
+        s8 = src + (y+row) * src_stride + x * 4;
+        hash = wyhash((const void*)s8, 64 * 4, hash, _wyp);
+    }
+    return hash;
```
matt335672, line 608: Small point (and I haven't looked at the generated code), but you could make the loop tighter with:-

```c
    s8 = src + (y * src_stride) + (x * 4);
    for(row = 0; row < 64; row++)
    {
        hash = wyhash((const void*)s8, 64 * 4, hash, _wyp);
        s8 += src_stride;
    }
```

metalefty: Indeed, it might be optimized by the compiler though.

metalefty:
I tested both wyhash and xxhash64 with lazy color convert. See #300 for results. I'm inclined to use wyhash.

@jsorg71 @Nexarian @matt335672 Any thoughts about which hash algorithm to use?

jsorg71:
I'll agree meta, I feel like we vetted wyhash well.  I'm not concerned about the author as this is not used for security.

matt335672:
I think you've done a great job looking at this @metalefty. Because wyhash has been tested already, that would be my preferred option. You've also shown that replacing this in the future is straightforward, if we need to.

metalefty:
Merging this.

PR #302, opened 2024-04-03
Switch to xxhash32/xxhash64 from CRC for capture tile diff

metalefty:
This is a draft PR to confirm which hash algorithm is faster to use in xorgxrdp.


metalefty:
Closing this. Perhaps we'll switch to xxhash in the future but we decided to use wyhash at themoment.

PR #309, opened 2024-04-30
simd rfx yuv

jsorg71:
In my tests, this speeds up the rgb to yuv 5+ x


matt335672 (approved):
I'm not qualified to comment on the assembler, but the plumbing looks OK to me.

metalefty:
The same from me.

PR #312, opened 2024-05-14
xorgxrdp capture changes for GFX h264

jsorg71:


module/rdpCapture.c
```diff
@@ -1187,14 +1407,16 @@ rdpCapture(rdpClientCon *clientCon, RegionPtr in_reg, BoxPtr *out_rects,
...
+        case 2: /* surface command RFX */
+            /* FALLTHROUGH */
+        case 4: /* GFX progressive */
             return rdpCapture2(clientCon, in_reg, out_rects, num_out_rects, id);
-        case 3:
-        case 5:
+        case 3: /* surface command h264 */
             /* used for even align capture */
             return rdpCapture3(clientCon, in_reg, out_rects, num_out_rects, id);
+        case 5: /* GFX h264 */
```
Nexarian: The only thing I might change is now that we have 5 different capture modes, defining them in a clearer enumeration with real names might be in order. It's really confusing for those who aren't ultra-familiar with the code to understand what these mean.

jsorg71: Sure, that makes sense.  I guess the function names can change too so they are not called rdpCapture0 to rdpCapture5.

Nexarian: Yeah, now is the time to update them!

Nexarian: Note this requires a change in XRDP too, so I think the enumeration should be defined in the shared header between XRDP and XORGXRDP.

jsorg71: I did this, it seems to have broke CI for xorgxrdp but take a look

PR #313, opened 2024-05-15
pkg-config workaround for missing PKG_CHECK_VAR

jsorg71:


matt335672:
Looks OK, but this is also worth a read:-

https://github.com/p11-glue/p11-kit/issues/200

PR #303, opened 2024-04-03
Add mappings for media keys

sasha0552:
Related: neutrinolabs/xrdp#3022
Fixes: neutrinolabs/xrdp#1725
Fixes: #211

PR #316, opened 2024-06-12
Move to evdev keycodes

### LINKED ISSUE #211: XRDP Keyboard evdev support?
Is it possible to switch XRDP to evdev keyboard rules or at least make it compatible?

All ubuntu 20.04 flavors (gnome, xfce, ...) seem to use 'evdev' keyboards:
> setxkbmap -query -v
> Trying to build keymap using the following components:
> keycodes:   evdev+aliases(qwertz)
> types:      complete
> compat:     complete
> symbols:    pc+de+inet(evdev)
> geometry:   pc(pc105)
> rules:      evdev
> model:      pc105
> layout:     de

However when i connect to this same machines via XRDP i get a 'base' rule with a 'xfree86' setting:
> setxkbmap -query -v
> Trying to build keymap using the following components:
> keycodes:   xfree86+aliases(qwertz)
> types:      complete
> compat:     complete
> symbols:    pc+de
> geometry:   pc(pc104)
> rules:      base
> model:      pc104
> layout:     de

As far as i understand this might be intended to make it compatible for Xvnc sessions. However i don't use Xvnc sessions as i connect to Xorg sessions via xorgxrdp extensions.
So, how can i switch XRDP to evdev to make it behave exactly like it does on the X console?

My problem arises with certain applications (mostly remote consoles) that do work on the machine's X console as intended but having keyboard issues when i access my machine via XRDP. They all suffer the same problem of ignoring alt-gr keys (german keyboard @, \, ...) and actually no cursor keys - only cursor-down seems to be enter.
The same effect appears if i switch keyboard to 'evdev' in running XRDP session via `setxkbmap -rules evdev`
Then all applications have this same problem. So it seems to me that these applications won't play with 'base' mode but need 'evdev' rules?

So question is, what can be done to make XRDP play with 'evdev' rules. I tried this on ubuntu 20.04, xubuntu20.04 with xrdp 0.9.12 (ubuntu repos) and 0.9.18 built from source - all the same. Any ideas? 

Is this a bug or a limitation of the current implementation?
https://github.com/neutrinolabs/xrdp/issues/179 seems to imply that there is a way to switch to evdev rules (breaking Xvnc - but that would be ok for me) but it does not tell what needs to be done to achieve this. As this is quite old i'm not sure if it still applies.

matt335672:
Fixes #211

Depends on neutrinolabs/xrdp#3138

See also #303 

Uses the evdev X11 keycodes passed over from xrdp, so that these do not have to be mapped within xorgxrdp as well.

I'm hoping this is more readable than it was. I've removed a lot of magic numbers

### DIFF
--- a/module/rdp.h
+++ b/module/rdp.h
@@ -130,14 +130,24 @@ typedef struct _rdpPointer rdpPointer;
 
 struct _rdpKeyboard
 {
-    int pause_spe;
-    int ctrl_down;
-    int alt_down;
-    int shift_down;
+    /**
+     * State of tab key
+     *
+     * Used to remove mstsc.exe tab KeyRelease before and after TS_SYNC_EVENT
+     */
     int tab_down;
-    /* this is toggled every time num lock key is released, not like the
-       above *_down vars */
-    int scroll_lock_down;
+
+    /**
+     * Whether or not to skip the next numlock key press/release
+     */
+    int skip_numlock;
+
+    int x11_keycode_caps_lock; ///< Used in TS_SYNC_EVENT processing
+    int x11_keycode_num_lock; ///< Used in TS_SYNC_EVENT processing
+    int x11_keycode_scroll_lock; ///< Used in TS_SYNC_EVENT processing
+
+    int scroll_lock_down; ///< Whether key is up/down
+    int scroll_lock_state; ///< Toggle state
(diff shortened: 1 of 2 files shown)


matt335672:
Status update - this is turning into more work than I expected on the xrdp side, but I'm still looking at it.

Issue #320, opened 2024-07-16
black screen on login (NVidia / glamor)


utrobinmv:
### xrdp version

0.9.21.1

### Detailed xrdp version, build options

~~~
Configure options:
      --enable-ipv6
      --enable-jpeg
      --enable-fuse
      --enable-rfxcodec
      --enable-opus
      --enable-painter
      --enable-vsock
      --build=x86_64-linux-gnu
      --prefix=/usr
      --includedir=${prefix}/include
      --mandir=${prefix}/share/man
      --infodir=${prefix}/share/info
      --sysconfdir=/etc
      --localstatedir=/var
      --disable-silent-rules
      --libdir=${prefix}/lib/x86_64-linux-gnu
      --libexecdir=${prefix}/lib/x86_64-linux-gnu
      --disable-maintainer-mode
      --disable-dependency-tracking
      --with-socketdir=/run/xrdp/sockdir
      build_alias=x86_64-linux-gnu
      CFLAGS=-g -O2 -ffile-prefix-map=/build/reproducible-path/xrdp-0.9.21.1=. -fstack-protector-strong -Wformat -Werror=format-security 
      LDFLAGS=-Wl,-z,relro -Wl,-z,now -Wl,--as-needed
      CPPFLAGS=-Wdate-time -D_FORTIFY_SOURCE=2 
      PKG_CONFIG_PATH=/build/reproducible-path/xrdp-0.9.21.1/pkgconfig

  Compiled with OpenSSL 1.1.1w  11 Sep 2023
~~~

### Operating system & version

openmediavault_6.5.0-amd64 / Debian 11

### Installation method

dnf / apt / zypper / pkg / etc

### Which backend do you use?

xorgxrdp

### What desktop environment do you use?

Xfce

### Environment xrdp running on

a physical machine, 2 * nvidia rtx 3090 and 1 * nvidia rtx 4060ti

### Area(s) with issue?

Authentication, Graphic glitches

### Steps to reproduce

apt install -y xrdp xorgxrdp xfce4
apt install -y cuda
xfreerdp  /size:1600x850 /v:192.168.45.70 /uuser /p:password /sound /home-drive /clipboard

I have been using the openmediavault 6 build and xrdp packages for a very long time. However, recently I started having problems after installing cuda on this machine. After installing cuda, the rdp session stops working xorgXrdp. I realized that the problem is not with xrdp, But maybe you can help me solve the problem?

### ✔️ Expected Behavior

[...]

PR #322, opened 2024-07-18
glamor, check list of support devices

jsorg71:
This will help avoid issue like #320


matt335672:
LGTM - other devices can easily be added as required, either by us or by the user.

PR #323, opened 2024-07-19
support NULL cursors and large mono cursors

jsorg71:
Mono cursors larger than 32x32 were getting clipped, now up to 96x96 should work.
NULL or empty cursors didn't work, now they should.  Movie players or youtube can be used to test.


matt335672:
LGTM

PR #333, opened 2024-08-12
More evdev keyboard changes

### LINKED ISSUE #1721: i3wm wizard crashes part way through vis xRDP session
When starting for the first time the i3 window manager runs a wizard that generates a config file for further use.

This wizard fails to create a complete config when invoked via an xRDP session, demonstrating the same behaviour as described here: https://github.com/i3/i3/issues/2141#issuecomment-198514698.

Elsewhere in that issue this is attributed to an `xkb_keymap_key_get_syms_by_level returned no symbols for keycode` error. However the keys that i3 is looking for (the arrow keys) work fine in normal usage. In fact when the file is created correctly (eg by running i3-wizard locally) the now complete config works fine in an xRDP session.

Perhaps related is a behaviour where the media keys on my keyboard are being captured as letter keys, for example pressing mute types the letter "d". This is also confirmed by xev as the same code as pressing the d key.

matt335672:
Fixes neutrinolabs/xrdp#1721

This commit removes some unused code from the keyboard driver, and fixes a rare race condition between the Xorg server starting and the first RDP client connecting (see neutrinolabs/xrdp#1721)

Keyboard layout logging is also somewhat improved. On X server startup, a line like this is logged:-

```
[  4800.882] rdpLoadLayout: rules="evdev" model="pc104" variant=""layout="us" options=""
```

Following a client connect, the change is logged:-

```
[  4814.777] rdpLoadLayout: rules="evdev" model="pc104" variant=""layout="gb" options=""
```

### DIFF
--- a/xrdpkeyb/rdpKeyboard.c
+++ b/xrdpkeyb/rdpKeyboard.c
@@ -65,6 +65,11 @@ xrdp keyboard module
 #define LLOGLN(_level, _args) \
     do { if (_level < LOG_LEVEL) { ErrorF _args ; ErrorF("\n"); } } while (0)
 
+/* A few hard-coded evdev keycodes (see g_evdev_str) */
+#define CAPS_LOCK_KEY_CODE 66
+#define NUM_LOCK_KEY_CODE 77
+#define SCROLL_LOCK_KEY_CODE 78
+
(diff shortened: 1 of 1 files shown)


jsorg71:
LGTM

PR #337, opened 2024-08-15
[V0.10] set the mm in randr

matt335672:
Allows RandR to see the physical display size, if passed from the client.

See neutrinolabs/xrdp#3200 for background.

@metalefty - do you want this in V0.10?

(cherry picked from commit aaa06d2f933db714e331d07266405506e10d2cd3)


metalefty:
Yes, let's ship this to v0.10.

metalefty (approved):

PR #347, opened 2024-09-26
Use different frame rates for H.264 and RFX

metalefty:
An enterprise user requests this. Some users in the company report that current 25Hz frame rates (40ms between frames) are noticeably laggy even when clients and servers are placed in under a 1ms network. Not all people feel laggy but some people are sensitive to screen latency. 

I also feel 25Hz is noticeably laggy when network RTT >30-40ms.

Now we have H.264, a video codec which takes interframe differences.  So, a higher frame rate does not necessarily consume more network bandwidth. In contrast, RFX doesn't take interframe differences. 2x frame rates might be twice the network bandwidth. This is the reason for using different frame rates per codec.

I'm going to make the frame interval value configurable. Where's the best place to put the configuration? xorg.conf?  client_info and pass from xrdp?

@matt335672 Let me know your opinion.

### DIFF
--- a/module/rdp.h
+++ b/module/rdp.h
@@ -292,6 +292,8 @@ struct _rdpRec
     CARD32 last_event_time_ms;
     CARD32 last_wheel_time_ms;
 
+    CARD32 msFrameInterval;
+
     int conNumber;
 
     struct _rdpCounts counts;
--- a/module/rdpClientCon.c
+++ b/module/rdpClientCon.c
@@ -61,7 +61,6 @@ Client connection to xrdp
 #define USE_MAX_OS_BYTES 1
 #define MAX_OS_BYTES (16 * 1024 * 1024)
 
-#define MIN_MS_BETWEEN_FRAMES 40
 #define MIN_MS_TO_WAIT_FOR_MORE_UPDATES 4
 
 /*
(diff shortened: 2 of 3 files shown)


matt335672:
It's a quite complex question.

The point you make about different users having different sensitivities is interesting. That would suggest having a per-client configuration, but there's nothing in the specification to allow for that.

I think the best place to put the configuration (for now) may be in the `[Xorg]` section of `xorg.ini`, and then to pass it to xorgxrdp. That gives sites a bit more flexibility to set up different configurations for different users. It's easier to put it the configuration in `xorg.conf`, but much harder to change if we do that.

Code so far looks fine, BTW.

metalefty:
I think they provide separate VM instances for xrdp server per user. Therefore, it is not necessary to have a configuration for each client. It would be nice to have such a per-client configuration but high framerate also satisfies non-sensitive users.

> I think the best place to put the configuration (for now) may be in the [Xorg] section of xorg.ini, and then to pass it to xorgxrdp. 

Do you mean command line arguments of Xorg (xorgxrdp)?

matt335672:
I was thinking more along the lines of  (for example):-

<pre>
;
; Session types
;

[Xorg]
name=Xorg
lib=libxup.so
username=ask
password=ask
port=-1
code=20
# Needed for 0.9.x
ip=127.0.0.1
#keycode_set=base
<b>MinH264Refresh=16
MinRfxRefresh=32
</b>
</pre>

When we connect to xorgxrdp we can send the values over. That lets the users play with different values without restarting the session.

metalefty:
@matt335672 Can you have a look at this and https://github.com/neutrinolabs/xrdp/pull/3318?

matt335672:
Will do tomorrow (UK time).

matt335672:
LGTM!

PR #354, opened 2024-11-20
add support for nvenc and accel_assist

jsorg71:


matt335672:
This looks fine to me for an initial merge.

Am I correct in thinking that with these changes we can build with `--enable-lrandr` and it wont be used at run-time for the non-NVidia case?

jsorg71:
I created a page https://github.com/jsorg71/xorgxrdp/wiki/NVidia-xorgxrdp-and-RandR to try to explain the RandR status when using nvidia_drv.so with xorgxrdp.

matt335672:
@jsorg71 - we seem to have a case of crossing commits here.

I've merged #369, and it's now failing here:-

```
CC       rdpMain.lo
rdpMain.c:54:10: fatal error: xacestr.h: No such file or directory
   54 | #include <xacestr.h>
      |          ^~~~~~~~~~~
compilation terminated.
make[3]: *** [Makefile:697: rdpMain.lo] Error 1
```

The reason for this appears to be this commit:-

https://gitlab.freedesktop.org/xorg/xserver/-/commit/22a3014c50d0856b5542f184a9c147264f23ce79

I don't know enough about this area to know what the best thing to do to resolve this is.

jsorg71:
Looks like the test is building xserver 21.1.99.1.  That file, xacestr.h should only be included if the xserver is < 21.1.4.  Either the CI build job is pulling in the wrong xorg-server.h or this test is not working
```
#if XORG_VERSION_CURRENT < XORG_VERSION_NUMERIC(21, 1, 4, 0, 0)
#define XACE_DISABLE_DRI3_PRESENT
#endif
```

Where do you see the error?  All looks good at https://github.com/neutrinolabs/xorgxrdp/actions/runs/13416597185/job/37478810488?pr=369

matt335672:
Thanks Jay.

The error is here, and occurred after I merged in the build against the latest xorg:-

https://github.com/neutrinolabs/xorgxrdp/actions/runs/13519937988

It's the test that's failing. `XORG_VERSION_CURRENT` is set here:-

https://gitlab.freedesktop.org/xorg/xserver/-/blob/master/include/meson.build#L6-7
https://gitlab.freedesktop.org/xorg/xserver/-/blob/master/include/meson.build#L126

Which for `21.1.99.1` produces `12,101,099`

`XORG_VERSION_NUMERIC(21, 1, 4, 0, 0)` is translated to `210,104,000` here:-

https://gitlab.freedesktop.org/xorg/xserver/-/blob/master/hw/xfree86/common/xorgVersion.h?ref_type=heads#L36-37

Looking at other uses of `XORG_VERSION_NUMERIC` in our code, I think this test should be against
`XORG_VERSION_NUMERIC(1, 21, 1, 4, 0)` which maps to `12,101,004` if my maths is correct.

I'll put a PR together to test the CI.

jsorg71:
oh wow, I've been using that macro wrong.  I better check the other places it's used.

matt335672:
I've been talking to @metux about a couple of other calls which you might want to be aware of.

Thread is here:-

https://gitlab.freedesktop.org/metux/xserver/-/commit/57aab7c9c3761e8c8a8bb1ba0f25fec26799aebd

I'll work on getting a CI going against the incubator branch so we catch these things sooner rather than later.

Issue #362, opened 2025-02-04
Workaround for missing "display" var (removed from xserver)


Luckylazuli:
Hi, this is my first post in GitHub, so sorry if I do something wrong.

I noticed that on my openSUSE Distro xrdp had a major update and wasn't working anymore, since also, I build xserver devel from git.
I don't know exactly if this is a common issue or specific for my system. Anyhow, I also build xrdp and finally found out, that xorgxrdp was missing "display" for the Xorg. Building xorgxrdp caused the compile error as excepted.

So I edited the code of '/module/rdpClientCon.c' - I gave up to understand if "display" looking for a existing socket or whatever.

Instead I made a simple workaround by checking for a set and customizable 'CUSTOM_XORGXRDP_DISPLAY' env var, with a fallback to "display = 0" if not set. Maybe this helps, idk:

```C
/******************************************************************************/
int
rdpClientConInit(rdpPtr dev)
{
    int i;
    char *ptext;
    char *endptr = NULL;
    const char *socket_dir;

    socket_dir = g_socket_dir();
    if (!g_directory_exist(socket_dir))
    {
        if (!g_create_dir(socket_dir))
        {
            if (!g_directory_exist(socket_dir))
            {
                LLOGLN(0, ("rdpClientConInit: g_create_dir(%s) failed", socket_dir));
                return 0;
            }
        }
        g_chmod_hex(socket_dir, 0x1777);
    }

// ------------------------------------------------------------------------------------------------------------
// 'display' member variable in 'opaque.h' of xserver removed, workaround by customizable environment variable
    const char *display = getenv("CUSTOM_XORGXRDP_DISPLAY");
    if (display == NULL) {
        LLOGLN(0, ("Env var 'CUSTOM_XORGXRDP_DISPLAY' not set, fallback to display '0'."));
        display = "0";
    }
// ------------------------------------------------------------------------------------------------------------

    errno = 0;
    i = (int)strtol(display, &endptr, 10);
    if (errno != 0 || display == endptr || *endptr != 0)
    {
        LLOGLN(0, ("rdpClientConInit: can not run at non-interger display"));
        return 0;
    }

    /* TODO: don't hardcode socket name */
    g_sprintf(dev->uds_data, "%s/xrdp_display_%s", socket_dir, display);
    if (dev->listen_sck == 0)
    {
        unlink(dev->uds_data);
        dev->listen_sck = g_sck_local_socket_stream();
[...]

PR #366, opened 2025-02-12
Code improvements

### LINKED ISSUE #365: Xorg alignment review
The xnf*alloc*() macros shoudn't be used anymore, they're going away soon. 
See xorg header "os.h" for the corresponding API functions.

OTOH, it's IMHO better to fail gracefully instead of crashing the whole Xserver.

https://github.com/neutrinolabs/xorgxrdp/blob/9698ba0be46cb692211c6c68b91720fbcc1f2bdb/xrdpdev/xrdpdev.c#L131

matt335672:
Fixes #365 (Xorg alignment review)

Suggested by the xorg team for future compatibility.

~~Draft for now, as I haven't run up and tested. I also need to check compatibility with older X servers for the removal of `xnfcalloc()`~~

Compiles OK against latest version of Xorg (gitlab commit 2e0c19b6d93663c6874d133f8a88267eee8cc16e)  when #362 issues are taken into account:-
1) Missing `display`
2) Missing `miPointerInitialize()`

Because of the last of these I can't test against latest ATM.

### DIFF
--- a/xrdpdev/xrdpdev.c
+++ b/xrdpdev/xrdpdev.c
@@ -127,8 +127,8 @@ rdpAllocRec(ScrnInfoPtr pScrn)
     {
         return TRUE;
     }
-    /* xnfcalloc exits if alloc failed */
-    pScrn->driverPrivate = xnfcalloc(sizeof(rdpRec), 1);
+    /* XNFcallocarray exits if alloc failed */
+    pScrn->driverPrivate = XNFcallocarray(sizeof(rdpRec), 1);
     return TRUE;
 }
 
@@ -149,8 +149,8 @@ rdpFreeRec(ScrnInfoPtr pScrn)
 static Bool
 rdpPreInit(ScrnInfoPtr pScrn, int flags)
 {
-    rgb zeros1;
-    Gamma zeros2;
+    rgb zeros1 = {0};
+    Gamma zeros2 = {0};
     int got_res_match;
 #if XORG_VERSION_CURRENT < XORG_VERSION_NUMERIC(1, 16, 0, 0, 0)
     char **modename;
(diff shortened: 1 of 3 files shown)


matt335672:
Regarding the use of XNFcalloc vs xnfcalloc, it looks like the macro and function were both in XFree86 4.3.0.1 (Nov 2003):-

https://gitlab.freedesktop.org/xorg/xserver/-/blob/xf86-4_3_0_1/include/os.h#L83

Since this pre-dates our oldest X server version currently in use (on Ubuntu 16.04 LTS) by over a decade, I think we can comfortably state that not using the macro won't cause any problems. The compilation is fine on Ubuntu 16.04 LTS, by the way.

matt335672:
Following https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/1787 I've tested this on the latest Xorg/xserver master branch, and on Ubuntu 22.04.

Regarding [this comment](https://github.com/neutrinolabs/xorgxrdp/issues/365#issuecomment-2653192870) by @metux, I've not initialized fields which were previously only in the initializer-list as placeholders. For example:-

```patch
_X_EXPORT DriverRec g_DriverRec =
{
-    XRDP_VERSION,
-    g_xrdp_driver_name,
-    rdpIdentify,
-    rdpProbe,
-    rdpAvailableOptions,
-    0,
-    0,
-    rdpDriverFunc
+    .driverVersion = XRDP_VERSION,
+    .driverName = g_xrdp_driver_name,
+    .Identify = rdpIdentify,
+    .Probe = rdpProbe,
+    .AvailableOptions = rdpAvailableOptions,
+    .driverFunc = rdpDriverFunc
};
```

The two `0` values in the old initializer-list correspond to members `pointer` and `refCount`. These will be initialized to NULL and 0 respectively with the new list. From 6.7.8 of ISO/IEC 9899:TC3 (emphasis mine):-

> 19 ) The initialization shall occur in initializer list order, each initializer provided for a particular subobject overriding any previously listed initializer for the same subobject; <b>all subobjects that are not initialized explicitly shall be initialized implicitly the same as objects that have static storage duration.</b>

metux:
> Regarding [this comment](https://github.com/neutrinolabs/xorgxrdp/issues/365#issuecomment-2653192870) by @metux, I've not initialized fields which were previously only in the initializer-list as placeholders. For example:-

Exactly. Actually, I'd advice to do so. That makes the code easier to maintain :) 
There only exception might be if you want to explicitly tell something to the human reader of the source code (eg. you might set the minor version explicitly, even if it's zero)

jsorg71 (approved):
Look good to me

Issue #340, opened 2024-08-23
Unable to compile 


Atra99:
I'm attempting to compile xorgxrdp in Manjaro Linux. When running make, I get the following error.

```
rdpClientCon.c: In function 'rdpClientConInit':
rdpClientCon.c:1564:21: error: 'display' undeclared (first use in this function)
1564 |     i = (int)strtol(display, &endptr, 10);
     |                     ^~~~~~~
rdpClientCon.c:1564:21: note: each undeclared identifier is reported only once for each function it app
ears in
make[3]: *** [Makefile:681: rdpClientCon.lo] Error 1
```
I'm sure I'm doing something wrong, but I can't find any information online about what would be causing this

matt335672:
It's a bit obscure.

The variable `display` is a global X server variable defined in `/usr/include/xorg/opaque.h` on my (Ubuntu) system. That file is pulled in by `#include <xf86_OSproc.h>` Maybe there's another `opaque.h` on your include path(?)

k9dkq:
I'm having the same issue on Gentoo compiling the latest version.  I checked `/usr/include/xorg/opaque.h` and found it did not include a global variable for 'display'.

matt335672:
I don't have a Gentoo install here, but it looks like this might be the problem:-

https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/1351

The symbols are now in dix/dix_priv.h. You could try including this for now and see if it fixes the immediate issue. Please report back regardless, and we'll try to find a better solution.

k9dkq:
It still didn't find "display" defined in the include chain.

matt335672:
I don't have a Gentoo installation to hand, and this version of `opaque.h` hasn't made it to Arch/Manjaro yet (which I do have to hand). I've got 21.1,14 on Manjaro which doesn't have this change.

@Atra99 - did you get to the bottom of the problem on your Manjaro installation?
@k9dkq - can you take a patch? If so, which version of xorgxrdp are you trying to compile?

k9dkq:
I'm compiling against the latest xrdp-0.10.1 and xorgxrdp-0.2.15.  I found the the "display" global variable in dix/dix_priv.h as you specified but I'm had clone the xserver git repository.  I'm guessing I'll have to change the Makefile to specify the location of the xserver source.

jsorg71:
I have noticed the Xorg guys are removing a bunch of exported symbols.  They claim they are not used by anyone.
[...]

PR #369, opened 2025-02-17
Add xorg sources to CI

matt335672:
Created mainly for discussion at this point. I've got to a point where I need some input to make this workable.

This PR has been prompted by #340 and #362.

The PR adds a script which runs as part of the standard CI to make a basic build of the xorgxrdp module against the latest master from https://gitlab.freedesktop.org/xorg/xserver

The script has been tested on Ubuntu 22.04, 24.04 and Debian testing

I've got a couple of questions which I'd like anyone (including @metux) to comment on:-
1) Is the checkin CI the right place for this, or should we be doing a nightly or weekly run, like the Coverity scan for xrdp? 
2) Is a basic build OK, or should we be going for as comprehensive a build as possible?

Personally I'm thinking a nightly or weekly run would be better, but a checkin is fine, in which case I need to add some caching to this.

### DIFF
--- a/scripts/build_latest_xorg_apt.sh
+++ b/scripts/build_latest_xorg_apt.sh
@@ -72,6 +72,10 @@ PYTHON_VENV="$SOURCE_DIR/python"            ; # Used for meson/ninja
 # The build directory (set later)
 BUILD_DIR=
 
+# Dependencies within the BUILD_DIR. When we know the BUILD_DIR, it
+# is added as a prefix to these
+BUILD_TARGET=lib/pkgconfig/xorg-server.pc    ; # What we are trying to make
+
 # ------------------------------------------------------------------------------
 # C R E A T E   B U I L D   M O D F I L E
 #
@@ -182,6 +186,13 @@ if ! cd "$1" || ! [ -w . ]; then
     exit 1
 fi
 BUILD_DIR=$(pwd)
+BUILD_TARGET="$BUILD_DIR/$BUILD_TARGET"
+
+# Have we run this script before?
+if [ -e "$BUILD_TARGET" ]; then
+    echo "- Target $BUILD_TARGET exists. The script will not be run again" >&2
+    exit 0
+fi
 
 # Debian has an extra pkgconfig directory to consider in
 # lib/$(uname -p)-linux-gnu/, which meson from pip doesn't pick up on.
--- a/scripts/build_latest_xorg_apt.sh
+++ b/scripts/build_latest_xorg_apt.sh
@@ -1,0 +1,231 @@
+#!/bin/sh
+################################################################################
+# B U I L D   W I T H   L A T E S T   X O R G   A P T . S H
+#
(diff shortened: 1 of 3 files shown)


metalefty (approved):

scripts/build_latest_xorg_apt.sh
```diff
@@ -0,0 +1,231 @@
...
+if [ -z "$XDG_RUNTIME_DIR" ]; then
+    export XDG_RUNTIME_DIR=/run/user/$(id -u)
+fi
+AUTORESUME_FILE="$XDG_RUNTIME_DIR/xserver-autores.txt"
+
+# Our source dir. This is currently hard-coded
+SOURCE_DIR="$HOME/xserver-src"
+
+# Dependencies in the source dir
+BUILDER="$SOURCE_DIR/util/modular/build.sh" ; # xorg modular build script
```
metux, line 67: Where is this script coming from ?

metux:
> 1. Is the checkin CI the right place for this, or should we be doing a nightly or weekly run, like the Coverity scan for xrdp?

Since you wanna catch possibly breaking changes in Xorg upstream, a cyclic (eg. nightly) run seems plausible. Otherwise you'd only recognize problems when somebody creates a MR.

A good place for catching *new* things (that aren't in master yet), you can additionally test against my incubator branch:
https://gitlab.freedesktop.org/metux/xserver/-/tree/incubate?ref_type=heads

But be aware that this branch is unstable (and frequently rebased), so your test might create certain false alarms.

scripts/build_latest_xorg_apt.sh
matt335672: It's fetched here:-

https://github.com/neutrinolabs/xorgxrdp/blob/4e9b011e9bcff8affb77e65bb6a9e8cdc92cde1b/scripts/build_latest_xorg_apt.sh#L210-L213

This could be clearer though. I'll update the comment and add an additional check after the checkout.

matt335672:
Thanks @metux 

I can look at using your incubator, certainly. Is there a corresponding util/modular repo (or branch) I should be using with it?

matt335672:
I've added two extra commits to make the script a bit clearer, and also to make these changes to the CI:-
1) Xorg latest build is in a separate section
2) Xorg latest build now also uses Glamor
3) Xorg latest build is run up briefly with xorgxrdp module.
3) Latest build directory is cached for a particular branch to speed up CI for developers.

I think this is fine for regular commit CI. We can also add a daily/weekly CI against @metux's inubator as a separate PR.

metux:
> Is there a corresponding util/modular repo (or branch)

I don't think anybody at xorg is still using these ancient scripts.

matt335672:
I see!

I picked them up as it's the first option here:-

https://www.x.org/wiki/Building_the_X_Window_System/

Is there a better, more maintainable way to accomplish an X server build you would recommend?

PR #370, opened 2025-02-19
Add incubate capability to xorg build script

matt335672:
~~Draft PR, more to capture some ideas than anything else. No review required yet, but comments welcome~~

See #369 for the background.

There are two commits in this PR:-

1) Updates to build_latest_xorg_apt.sh:-

    1) Add -x option to build_latest_xorg_apt.sh<br>This option allows the Xorg server to be built from @metux's incubate branch (and maybe other sources)
    
    2) Add extra freedesktop packages needed to build the incubate branch.<br>The packages libxcb-util and libxcb-wm are needed to build the current incubate branch.
    
    3) Remove hard-coded SOURCE_DIR<br>This allows two branches to be stored on the same development machine. The SOURCE_DIR is now dependent on the BUILD_DIR
    
     4) Don't run sudo unless we need to<br>This checks whether OS packages are installed before using  'sudo apt install' to install them. This is intended for interactive use.

2) Add a github workflow to run the script every Sunday, and also on demand.

An example of the results of a manual run of the script can be found [here](/matt335672/xorgxrdp/actions/runs/13566596152)

### DIFF
--- a/.github/workflows/incubate.yml
+++ b/.github/workflows/incubate.yml
@@ -1,0 +1,34 @@
+name: Incubate
+
+on:
+  schedule:
+    - cron: "0 0 * * 0"
+  workflow_dispatch:
+
+jobs:
+  build_with_xorg_incubate:
+    name: Build with xorg Incubate branch
+    runs-on: ubuntu-latest
+    env:
+      CC: gcc
+      CONF_FLAGS: --enable-glamor
+      XRDP_CFLAGS: -I${{ github.workspace }}/xrdp/common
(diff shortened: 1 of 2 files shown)


matt335672:
I'm going to merge this now so I can test it runs this weekend as expected.

matt335672:
Here's an example of the usefulness of this facility:-

https://gitlab.freedesktop.org/metux/xserver/-/commit/57aab7c9c3761e8c8a8bb1ba0f25fec26799aebd#note_2800041

PR #376, opened 2025-03-19
Update autoconf scripts from 2.65 to 2.69

matt335672:
Tested on Ubuntu 16.04 (autoconf 2.69) and Ubuntu 24.04 (autoconf 2.71)

2.69 is not the latest version, but it is the version used by xrdp.

Various obsolete macros have been updated. In addition:-
1) The file m4/nasm.m4 has been renamed to m4/axrdp_prog_nasm.m4, and the private macro it contains has been renamed from `AC_PROG_NASM` to `AXRDP_PROG_NASM`. This is because the AC_* namespace is reserved for autoconf macros.
2) The m4/ax_*.m4 library files have been update to the latest versions from https://git.savannah.gnu.org/cgit/autoconf-archive.git
3) Following 2), missing macros `AX_COMPILER_VENDOR` and `AX_PREPEND_FLAG` have also been added.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1,6 +1,6 @@
 # Process this file with autoconf to produce a configure script
 
-AC_PREREQ(2.65)
+AC_PREREQ(2.69)
 # package version must be x.y.z
 AC_INIT([xorgxrdp], [0.10.80], [xrdp-devel@googlegroups.com])
 package_version_major=$(echo ${PACKAGE_VERSION}|cut -d. -f1)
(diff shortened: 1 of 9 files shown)


matt335672:
@metalefty - do you want a backport to v0.10.x, or would you prefer to stay on 2.65 for the lifetime of v0.10.x?

metalefty:
Wow, 2.65 is an older version that is included in Ubuntu 16.04. Let's switch to 2.69.

Issue #342, opened 2024-08-27
i-beam cursor switches to black square when typing


jkeltz:
I have been running latest xrdp and xorgxrdp 0.9 for quite some time.
I compile from source.  I'm running Rocky Enterprise Linux 8.10 (RHEL8.10).
Today,  I updated to xorgxrdp 0.10.2 and xrdp 0.10.1 and am having one issue.

When I login to my XRDP session, my gnome session starts.  I open a terminal.   If I hit backspace, I hear the beep (because I've configured sound) and immediately the mouse cursor switches from the (I think it's called) "I-beam" to just a black square!   I can get it to switch back to the "I-beam" again by just moving the mouse.
If I start typing in the Terminal window, the cursor switches back to the black square, and again, if I move the mouse it switches back to the I-Beam.
Now, if I click on the terminal so that my text is going to go in there, and move the mouse cursor out of the terminal Window so that it turns to the arrow, then I can type, and my text will go into the window, and the cursor will not change...
So this seems to impact the I-beam cursor only...  But why? This has never been an issue for me in the past.

I have this installed on > 150 systems, and the same experience on all of them (unfortunately).

(I believe this is an xorgxrdp issue... My apologies if this is an xrdp issue).

jkeltz:
I should add that I viewed the xrdp log and see some errors that could be relevant:

```
[2024-08-27T15:58:51.685-0400] [INFO ] starting xrdp with pid 10427
[2024-08-27T15:58:51.686-0400] [INFO ] address [0.0.0.0] port [3389] mode 1
[2024-08-27T15:58:51.687-0400] [INFO ] listening to port 3389 on 0.0.0.0
[2024-08-27T15:58:51.688-0400] [INFO ] xrdp_listen_pp done
[2024-08-27T15:58:58.359-0400] [INFO ] Socket 14: connection accepted from 130.63.94.85:36454
[2024-08-27T15:58:58.361-0400] [INFO ] Using default X.509 certificate: /xconf/xrdp/cert.pem
[2024-08-27T15:58:58.362-0400] [INFO ] Using default X.509 key file: /xconf/xrdp/key.pem
[2024-08-27T15:58:58.363-0400] [INFO ] Security protocol: configured [SSL|RDP], requested [SSL|HYBRID|RDP], selected [SSL]
[2024-08-27T15:58:58.368-0400] [INFO ] Connected client computer name: Guacamole RDP
[...]

PR #380, opened 2025-03-22
add system pointer

jsorg71:


jsorg71:
I think this may also fix #342

matt335672 (approved):
LGTM

PR #384, opened 2025-04-04
Use different struct from xrdp

matt335672:
In the proposed neutrinolabs/xrdp#3492,  xrdp now passes us `struct xorgxrdp_client_con`, rather than `struct xrdp_client_con`

This PR won't compile until neutrinolabs/xrdp#3492 is merged.


matt335672:
Superceded by #385

PR #395, opened 2025-10-29
Support DRI3 version 1.3

### LINKED ISSUE #345: support explicit modifiers
Currently xrdp only supports DRI3 v1.0. This means only implicit modifiers are supported, which is suboptimal for drivers that are unable to effectively communicate their modifier selection without the explicit API.

xrdp should support explicit modifiers with DRI3 v1.3.

kballen1:
Fixes #345 

Add dri3_screen_info struct version 2 fields to enable support for DRI3 v1.3, including support for explicit modifiers.

I should mention that I have very little familiarity with Xorg drivers and DRI, I just followed the pattern in the existing code. It would probably be best if someone who knows this area could make sure this looks good, but it has been working for me and it fixes a noisy DRI3 error from the android studio emulator.

### DIFF
--- a/xrdpdev/xrdpdri3.c
+++ b/xrdpdev/xrdpdri3.c
@@ -204,10 +204,10 @@ rdpDri3Init(ScreenPtr pScreen)
     rdp_dri3_info.get_modifiers = rdpDri3GetModifiers;
     rdp_dri3_info.get_drawable_modifiers = rdpDri3GetDrawableModifiers;
 #endif
-    LLOGLN(0, ("rdpScreenInit: rdp_dri3_info.version = %lu", (unsigned long)rdp_dri3_info.version));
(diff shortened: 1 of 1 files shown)


matt335672:
@jsorg71 - this is far more your area than mine. Can you take a look?

jsorg71:
It looks good, testing on a server now.

Nexarian:
@kballen1 I appreciate you doing this. In my day job I work in very very high level language land: AWS, Kubernetes, Python, and Java once upon a time. It was a huge lift to learn the XRDP driver level C, and I'm still nowhere near as good as @jsorg71 is.

The best way is to do exactly what you did: Try, play with it, and read lots and lots of code.

Keep it up!

jsorg71:
This has been working great.  Hopefully we can rebase as it a bit older but either way.

matt335672:
@metalefty - do you want this one backporting to v0.10?

PR #396, opened 2025-10-31
xinput: Use standard type name in InputInfoPtr

### LINKED ISSUE #394: XINPUT mouse / keyboard type vs Chrome pointer detection
Quoting the docs for [XListInputDevices](https://linux.die.net/man/3/xlistinputdevices):

```
           The type field is of type Atom and indicates the nature of the
           device. The type will correspond to one of the following strings
           (defined in the header file XI.h):

           XI_MOUSE XI_TABLET XI_KEYBOARD XI_TOUCHSCREEN XI_TOUCHPAD
           XI_BUTTONBOX XI_BARCODE XI_TRACKBALL XI_QUADRATURE XI_ID_MODULE
           XI_ONE_KNOB XI_NINE_KNOB XI_KNOB_BOX XI_SPACEBALL XI_DATAGLOVE
           XI_EYETRACKER XI_CURSORKEYS XI_FOOTMOUSE XI_JOYSTICK
```

(There are mentions on the [XOrg Input Driver Specifications](https://www.x.org/wiki/XOrgInputDriverSpec/) and [XInput Extension Specifications](https://www.x.org/wiki/XInputSpec/) about using HID names, but as those proposals appear to be old and incomplete, I'm doubtful whether they should be followed.)

Therefore, when we're looking for a mouse in the `XListInputDevices(3)` result, one approach would be to enumerate devices searching for `XI_MOUSE`. This is what Chrome Stable on Linux now appears to do [when detecting input devices](https://source.chromium.org/chromium/chromium/src/+/main:ui/events/platform/x11/x11_hotplug_event_handler.cc;l=423;drc=eacc76afbe1d2a055571fc8f66ab67d17530956c).

But xorgxrdp uses `"Mouse"` not `XI_MOUSE`. The result is that [when detecting hover capabilities](https://patrickhlauke.github.io/touch/pointer-hover-any-pointer-any-hover/) in the browser, Chrome will not use CSS `@media (hover: hover)` styling.

There is a recent [open bug against Chromium](https://issues.chromium.org/issues/443001333), so it may change the detection logic, but I wanted to ask here why the XINPUT-standard type Atom strings could be used, but aren't?

The same question arises for `g_Keyboard_str`.

metalefty:
We should use strings defined in XI.h for the type field here. This fixes pointer detection issue in Chromium, resolves #394.

See also:
- https://issues.chromium.org/issues/443001333

-----

Tested with Chromium 141.0.7390.122 (FreeBSD port).

### Before applying the fix (no pointers)
508020565-7dc531ea-ea22-4d08-ace9-c40ae56ac5fc

### After applying the fix (pointers are detected)
508094556-0262667d-fbba-4390-aa4d-fefe37c3c4fc

### DIFF
--- a/xrdpkeyb/rdpKeyboard.c
+++ b/xrdpkeyb/rdpKeyboard.c
@@ -70,7 +70,7 @@ static char g_evdev_str[] = "evdev";
 static char g_pc104_str[] = "pc104";
 static char g_us_str[] = "us";
 static char g_empty_str[] = "";
-static char g_Keyboard_str[] = "Keyboard";
+static char g_Keyboard_str[] = XI_KEYBOARD;
 
 static char g_xrdp_keyb_name[] = XRDP_KEYB_NAME;
 
@@ -443,7 +443,7 @@ rdpkeybPreInit(InputDriverPtr drv, IDevPtr dev, int flags)
     info->device_control = rdpkeybControl;
     info->flags = XI86_CONFIGURED | XI86_ALWAYS_CORE | XI86_SEND_DRAG_EVENTS |
                   XI86_CORE_KEYBOARD | XI86_KEYBOARD_CAPABLE;
-    info->type_name = "Keyboard";
+    info->type_name = g_Keyboard_str;
     info->fd = -1;
     info->conf_idev = dev;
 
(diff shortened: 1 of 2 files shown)


metalefty:
@jsorg71 Any thoughts?

rowlap (changes requested):
There's some slight code duplication we should address at rdpMouse.c line 405 for older X servers - `XORG_VERSION_NUMERIC(1, 9, 0, 1, 0)` although I'm not familiar with the versioning scheme.

metalefty:
Indeed, `XI_MOUSE` is also defined in `XI.h` via `xf86Xinput.h` in Xorg 1.8.x. We just need to replace them in the same way.

rowlap (approved):

matt335672 (approved):

PR #403, opened 2026-01-29
Add Xlibre support

metalefty:
Based on work by @b-aaz on [FreeBSD bug 291595](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=291595).

See also:
- https://github.com/b-aaz/xlibre-ports/commits/master/x11-drivers/xorgxrdp

I'll forward this to devel once merged.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -4,8 +4,16 @@ FreeBSD_task:
   env:
     XRDP_REPO: https://github.com/neutrinolabs/xrdp
     BUILD_TOOLS: git autoconf automake libtool pkgconf
-    BUILD_DEPS: nasm xorg-server
+    BUILD_DEPS: nasm $XORG_SERVER
     TEST_DEPS: xdpyinfo
+
+  matrix:
+    - name: FreeBSD build with Xorg
+      env:
+        XORG_SERVER: xorg-server
+    - name: FreeBSD build with Xlibre
+      env:
+        XORG_SERVER: xlibre-server
   prepare_script:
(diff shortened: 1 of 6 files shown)


module/rdpEgl.c
```diff
@@ -735,7 +735,11 @@ rdpEglCaptureRfx(rdpClientCon *clientCon, RegionPtr in_reg, BoxPtr *out_rects,
     {
         tmpval[0].val = GXcopy;
         tmpval[1].val = 0;
+#if defined(NullClient)
         ChangeGC(NullClient, rfxGC, GCFunction | GCForeground, tmpval);
+#else
+        ChangeGC(NULL, rfxGC, GCFunction | GCForeground, tmpval);
```
metalefty: Simply replacing NullClient with NULL works with Xorg, too. 

@matt335672  We might not need to check this macro. What do you think?

NullClient is defined like this in `include/dix.h`.
```
#define NullClient ((ClientPtr) 0)
```

matt335672: Another way might be to add this to one of our includes:

```c
#ifndef NullClient
#define NullClient ((ClientPtr) 0)
#endif
```

which I (personally) think is clearer.

matt335672 (approved):
Looks fine - I'm surprised that so little needs to be done.

One minor comment below.

module/rdpEgl.c
metalefty: Sounds good.

PR #404, opened 2026-01-30
CI: Bump FreeBSD version to 14.3

metalefty:


metalefty:
Oops, 15.0 is not available yet.

PR #411, opened 2026-02-24
Update logging

### LINKED ISSUE #410: xorgxrdp pushing absurd quantity of identical entries to systemd-journald
Hello,

I should first state that everything is working fine with my Xrdp setup. It's been a blessing to have a working RDP implementation because I hate VNC with a passion. Thank you for your work on this software. I use xrdp+xrdpxorg to provide me with remote access to an LXQt desktop on Arch Linux. Using xrdp 0.10.5-1 + xorgxrdp-glamor 0.10.4-1, and connecting from another Linux machine with Remmina.

I noticed that the journal on the server is inundated with these cryptic, identical entries:

```shell
Feb 11 13:50:11 <hostname> xrdp-sesman[1761]: rdpPolyArc:
Feb 11 13:50:11 <hostname> xrdp-sesman[1761]: rdpPolyArc:
Feb 11 13:50:11 <hostname> xrdp-sesman[1761]: rdpPolyArc:
Feb 11 13:50:11 <hostname> xrdp-sesman[1761]: rdpPolyArc:
Feb 11 13:50:11 <hostname> xrdp-sesman[1761]: rdpPolyArc:
Feb 11 13:50:11 <hostname> xrdp-sesman[1761]: rdpPolyArc:
Feb 11 13:50:11 <hostname> xrdp-sesman[1761]: rdpPolyArc:
Feb 11 13:50:11 <hostname> xrdp-sesman[1761]: rdpPolyArc:
Feb 11 13:50:11 <hostname> xrdp-sesman[1761]: rdpPolyArc:
Feb 11 13:50:11 <hostname> xrdp-sesman[1761]: rdpPolyArc:
Feb 11 13:50:11 <hostname> xrdp-sesman[1761]: rdpPolyArc:
Feb 11 13:50:11 <hostname> xrdp-sesman[1761]: rdpPolyArc:
```

These occur several hundreds of times per second when they occur (which is not all the time). Every now and then they'll stop for a bit. Then they may start again with this message:

```shell
Feb 11 13:50:30 <hostname> systemd-journald[771617]: [🡕] Suppressed 55472 messages from session-4.scope
```

You can see there was a 19-second gap between the stream above and the systemd-journald entry, after which it started churning out more of these rdpPolyArc entries.

Then they stop again and don't recur for several hours before picking up again. This tends to occur daily.

I'm counting over 15 million lines in journalctl since Feb 8 (worth about 4 GB under `/var/log/journal`) and the vast majority are these ones above. I'm not hard up for storage space but this feels like a silly number of entries to log and I don't what they mean since they say nothing except "rdpPolyArc". I looked it up, so now I understand what rdpPolyArc does, but it ends there.

I haven't figured out if there's a pattern i.e. whether something I'm doing is causing this. I had a passing thought that I have several screensavers on a random rotation in my LXQt desktop and that maybe some of them are causing this?

I looked at the timing of today's entries to compare it with when I was actively in Remmina vs. not using it, and they seem to occur either way. I always have an active X session on that box that I just connect to/disconnect from with Remmina. On a typical day I'll start Remmina on my client machine in the morning and it stays connected all day until I shut off the client machine. But I'll be switching from Remmina to other apps throughout the day and sometimes Remmina is unused for a while (but always connected to Xrdp).

So for today, I have a bunch of these entries from around 8:30am today when I was sleeping and Remmina was not connected to Xrdp. Then I have more of these entries around 9:45am and I was awake then, and Remmina was probably connected, but I don't know if I was using it at that time or doing something else.

Bottom line is it looks like this happens whether Remmina is connected or not, and whether I'm using it or not.

Anyway, is it possible to suppress these log entries? Regardless of their cause, I can't figure out what purpose they serve.

matt335672:
Fixes #410 

See #410 for the background to this. Thanks to @rowlap for some very useful comments. I'd appreciate more, @rowlap, if you have them!

Major features:
1. The `LLOGLN()` macro is now centralised in `rdpMisc.h`, rather than being source-file specific.
2. Some logging levels have been reduced to reduce noise
3. Messages are preceded with `(II)` or `(DB)`.
4. Previous behaviour can be achieved by starting `Xorg` with `-verbose 2 -logverbose 2`

The syntax of `LLOGLN()` is a little odd, and this is reflected in the new definition. I've checked the output of gcc under x86_64, and because constants are used, the compiler is able to choose the correct function to call at compile-time. Here is the code produced for the `rdpmousePlug()` function with `-g -O2`:

```
1:(gdb) disassemble/s
2:Dump of assembler code for function rdpmousePlug:
3:rdpMouse.c:
4:448	{
5:=> 0x00007ffff74113a0 <+0>:	endbr64
6:
7:449	    LLOGLN(10, ("rdpmousePlug:"));
8:   0x00007ffff74113a4 <+4>:	push   %rbx
9:   0x00007ffff74113a5 <+5>:	xor    %eax,%eax
10:
11:448	{
12:   0x00007ffff74113a7 <+7>:	mov    %rdi,%rbx
13:
14:449	    LLOGLN(10, ("rdpmousePlug:"));
15:   0x00007ffff74113aa <+10>:	lea    0xc6d(%rip),%rdi        # 0x7ffff741201e
16:   0x00007ffff74113b1 <+17>:	call   0x7ffff7411170 <g_log_trace@plt>
17:
18:450	    xf86AddInputDriver(&rdpmouse, module, 0);
19:   0x00007ffff74113b6 <+22>:	mov    %rbx,%rsi
20:   0x00007ffff74113b9 <+25>:	xor    %edx,%edx
21:   0x00007ffff74113bb <+27>:	lea    0x2d7e(%rip),%rdi        # 0x7ffff7414140 <rdpmouse>
22:   0x00007ffff74113c2 <+34>:	call   0x7ffff7411260 <xf86AddInputDriver@plt>
23:
24:451	    return module;
25:   0x00007ffff74113c7 <+39>:	mov    %rbx,%rax
26:   0x00007ffff74113ca <+42>:	pop    %rbx
27:   0x00007ffff74113cb <+43>:	ret
28:End of assembler dump.
```

The correct function is called directly at line 16.

### DIFF
--- a/module/rdpCapture.c
+++ b/module/rdpCapture.c
@@ -55,10 +55,6 @@ capture
 #include <glamor.h>
 #endif
 
-#define LOG_LEVEL 1
-#define LLOGLN(_level, _args) \
-    do { if (_level < LOG_LEVEL) { ErrorF _args ; ErrorF("\n"); } } while (0)
-
 #define RGB_SPLIT(A, R, G, B, pixel) \
     A = (pixel >> 24) & UCHAR_MAX; \
     R = (pixel >> 16) & UCHAR_MAX; \
--- a/module/rdpClientCon.c
+++ b/module/rdpClientCon.c
@@ -62,10 +62,6 @@ Client connection to xrdp
 #include "rdpRandRGrid.h"
 #endif
 
-#define LOG_LEVEL 1
-#define LLOGLN(_level, _args) \
-    do { if (_level < LOG_LEVEL) { ErrorF _args ; ErrorF("\n"); } } while (0)
-
 #define LTOUI32(_in) ((unsigned int)(_in))
 
 #define USE_MAX_OS_BYTES 1
--- a/module/rdpClientCon.c
+++ b/module/rdpClientCon.c
@@ -224,7 +224,7 @@ rdpClientConGotConnection(ScreenPtr pScreen, rdpPtr dev)
     rdpClientCon *clientCon;
     int new_sck;
 
-    LLOGLN(0, ("rdpClientConGotConnection:"));
+    LLOGLN(10, ("rdpClientConGotConnection:"));
     clientCon = g_new0(rdpClientCon, 1);
     clientCon->shmemstatus = SHM_UNINITIALIZED;
     clientCon->updateRetries = 0;
@@ -378,7 +378,7 @@ rdpShutdownAccelAssist(rdpPtr dev, rdpClientCon *clientCon) {
     int index;
     int exit_code = 0;
 
-    LLOGLN(0, ("rdpShutdownAccelAssist:"));
+    LLOGLN(10, ("rdpShutdownAccelAssist:"));
     if (clientCon->accel_assist_pid <= 0)
     {
         return 0;
@@ -432,7 +432,7 @@ rdpClientConDisconnect(rdpPtr dev, rdpClientCon *clientCon)
 {
     int index;
 
-    LLOGLN(0, ("rdpClientConDisconnect:"));
+    LLOGLN(10, ("rdpClientConDisconnect:"));
 
     if (dev->idleDisconnectTimer != NULL && dev->idle_disconnect_timeout_s > 0)
     {
@@ -845,7 +845,7 @@ rdpClientConResizeAllMemoryAreas(rdpPtr dev, rdpClientCon *clientCon)
 
     enum shared_memory_status shmemstatus;
 
-    LLOGLN(0, ("rdpClientConResizeAllMemoryAreas:"));
+    LLOGLN(10, ("rdpClientConResizeAllMemoryAreas:"));
 
     // Updare the rdp size from the client size
     clientCon->rdp_width = width;
@@ -1350,7 +1350,7 @@ rdpClientConProcessMsgClientInfo(rdpPtr dev, rdpClientCon *clientCon)
     int bytes;
     int i1;
 
-    LLOGLN(0, ("rdpClientConProcessMsgClientInfo:"));
+    LLOGLN(10, ("rdpClientConProcessMsgClientInfo:"));
     s = clientCon->in_s;
     in_uint32_le(s, bytes);
     if (bytes > sizeof(clientCon->client_info))
@@ -1625,7 +1625,7 @@ static int
 rdpClientConGotControlConnection(ScreenPtr pScreen, rdpPtr dev,
                                  rdpClientCon *clientCon)
 {
-    LLOGLN(0, ("rdpClientConGotControlConnection:"));
+    LLOGLN(10, ("rdpClientConGotControlConnection:"));
     return 0;
 }
 
@@ -1634,7 +1634,7 @@ static int
 rdpClientConGotControlData(ScreenPtr pScreen, rdpPtr dev,
                            rdpClientCon *clientCon)
 {
-    LLOGLN(0, ("rdpClientConGotControlData:"));
+    LLOGLN(10, ("rdpClientConGotControlData:"));
     return 0;
 }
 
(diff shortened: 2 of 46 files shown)


rowlap:
- 80fa4af551bec043a917ea9343a3a0adeb6e2074 looks good
- 91553044db6b81dafe0a32a6b111b3137c698f26 is obviously good; wasn't this previously flagged by the compiler as uninitialized?
- 39df7407d2a38102ae4b473680f7b616ed81989b is great removal of duplication, although I didn't carefully eyeball all the deletions for similarity as I assume the `LLOGLN` macro is intended to be identical throughout the codebase
- babcb2a3ed588dd4af2b00f9427e454bec494fcd seems fine, but I can't spot the API docs for `LogVMessageVerb`

Just to confirm, the `(II)` or `(DB)` prefixes are from Xorg, reflecting the level, not present in any xorgxrdp code.

This looks like printf-style/variadic formatting, but there's no `%s` anywhere I can spot, which might be a [CWE-134](https://cwe.mitre.org/data/definitions/134.html) format string vulnerability. I don't know whether any data passing through `LLOGLN` is externally controlled, but perhaps worth tightening up. If this is a bug, it's also present in the current version.

Are there any locations where a user-controlled string is passed directly to `LLOGLN` without using `%s`? Sorry for the paranoia.

matt335672:
Thanks for that @rowlap 

1) the `(II)` and `(DB)` prefixes are provided by `LogVMessageVerb()`.
2) The compiler didn't previously pick up on the uninitialised variables in 91553044db6b81dafe0a32a6b111b3137c698f26. It may be doing so now because of the PRINTFLIKE macro. There is more work we could do separately in terms of adding some static analysis to the CI.
2) you're right in that `LogVMessageVerb()` isn't in the [DDX Design](https://www.x.org/releases/current/doc/xorg-server/ddxDesign.html) document, but then there are no functions in that document which take a `va_list`. To work around that, I'd need to call `xf86MsgVerb()` from `LLOGLN`, which makes the macro more complex, as it would need to work out the parameters for `xf86MsgVerb()`. IMO, we'd be better off in that case replacing all the calls to `LLOGLN` with explicit calls to `xf86MsgVerb()`. This is doable, but messy. Also, currently `xf86MsgVerb()` [calls LogVMessageVerb() directly](https://cgit.freedesktop.org/xorg/xserver/tree/hw/xfree86/common/xf86Helper.c#n1099), so in the event this stops working, we can hopefully find a similar function to call.
3) As regards the security question, that's fine; it's definitely the right time to ask such things.. There are 605 calls to `LLOGLN()` in the branch:
```
$ find . -name \*.c -exec grep LLOGLN {} + | wc -l
605
```
and all of those are followed by a static string as the first parameter:
```
$ find . -name \*.c -exec grep 'LLOGLN([01]*, ("' {} + | wc -l
605
```

What are your thoughts on the use of `LogVMessageVerb()`?

**Edit:** One way to use `xf86MsgVerb()` directly might be to output the message to a buffer within our code, but then we're doing work which may be thrown away anyway if the logging level isn't sufficient. The logging level is private within the logging module, so we can't query it.

rowlap:
To be pragmatic, the Xorg logging API is unchanging, and not many people work on xorgxrdp code, therefore this looks good to go. Nothing here precludes future improvements.

One small suggestion for readability would be to `#define` the levels in rdpMisc.h so they could be symbolic, not numeric. `LLOGLN_DEBUG` etc., but I'm nitpicking as they've been arbitrary integers for years without complaint. I think that would still allow the `g_log_*` inlining.

matt335672:
Thanks @rowlap

I'm happy to move to symbolic logging. This can be largely automated, but since it involves a visit to every `LLOGLN` invocation anyway, I'm tempted to go a bit further and make the logging similar to that used in the main product, eg:

```c
LLOGLN(0, ("This is an info message reason:%s", reason));
```

becomes:

```c
LOG(LOG_LEVEL_INFO, "This is an info message reason:%s", reason);
```

I'd appreciate your thoughts. I think that will be the subject of another PR. I'll merge this one.

@metalefty - do you want this one in v0.10? We can talk about symbolic logging later.

matt335672:
I've had a look at moving to the LOG() macro used by xrdp, and it's actually quite easy to automate. I've got something working and I hope to have a PR ready soon for discussion.

PR #413, opened 2026-02-26
Move to LOG() macro

matt335672:
~~Draft for now - I need to do more testing.~~ Comments welcome.

Following a suggesting from @rowlap, this PR replaces the `LLOGLN()` macro with the `LOG()` macro and levels from xrdp.

The move was largely automated, with some tweaking afterwards to fix indentation. A couple of logging levels have been changed to the new `LOG_LEVEL_WARNING`.

The use of symbolic names illustrates perfectly that logging in this module is currently all over the place!

This can be back-ported to v0.10.x if desired.


metalefty (approved):
LGTM, thanks you for your work!

PR #419, opened 2026-03-15
Optimize keyboard lock state synchronization

### LINKED ISSUE #259: xfreerdp: caps lock sometimes gets out of sync / works inverted
sometimes when i press caps lock, the state on the server doesn't change: i.e. the led indicator shows on but when i type all letters are upcase and vice versa.

**To Reproduce**
press caps lock.

**Expected behavior**
led indicator matches rdp state.

**Additional context**
to test i've used: watch -n 0.1 "xset q"

server + client is debian 12.

xrdp is v0.9.21.1-1

i reported the same to freerdp: https://github.com/FreeRDP/FreeRDP/issues/8946

tsz8899:
Fixes https://github.com/neutrinolabs/xorgxrdp/issues/259
Introduces a deferred synchronization mechanism that aligns Caps Lock and Num Lock states upon the user's first keyboard interaction.

- Added persistent state tracking for sync flags.
- Implemented explicit state matching (X11 vs RDP).
- Added support for re-synchronization on session reconnect.

Is this optimization plan feasible?

### DIFF
--- a/xrdpkeyb/rdpKeyboard.c
+++ b/xrdpkeyb/rdpKeyboard.c
@@ -69,6 +69,10 @@ static char g_empty_str[] = "";
 static char g_Keyboard_str[] = XI_KEYBOARD;
 
 static char g_xrdp_keyb_name[] = XRDP_KEYB_NAME;
+static int g_initial_sync_completed = 0;
+static int g_deferred_sync_flags = 0;
+/* Forward declaration for deferred synchronization */
+static void KbdSync(rdpKeyboard *keyboard, int param1);
 
 static int
 rdpLoadLayout(rdpKeyboard *keyboard, struct xup_client_info *client_info);
@@ -123,6 +127,14 @@ static void
 KbdAddEvent(rdpKeyboard *keyboard, int down, int param1, int param2,
             int param3, int param4)
 {
+
+    /* Perform deferred synchronization on the first keyboard interaction */
+    if (!g_initial_sync_completed && down)
+    {
+        g_initial_sync_completed = 1;
+        LOG(LOG_LEVEL_DEBUG, "KbdAddEvent: First key press, enforcing deferred sync");
+        KbdSync(keyboard, g_deferred_sync_flags);
+    }
     int x_keycode = param1;
     int rdp_scancode = SCANCODE_FROM_KBD_EVENT(param3, param4);
     int type = down ? KeyPress : KeyRelease;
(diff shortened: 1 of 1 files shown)


matt335672:
Thanks again @tsz8899 

A couple of comments:

1) Persistent state tracking isn't feasible over a disconnect/reconnect as you cannot guarantee that the same client will reconnect, or if it does, that it will have its modifier keys in the same state.
2) The relevant text from [MS-RDPBCGR] is sections [2.2.8.1.1.3.1.1.5](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/6c5d0ef9-4653-4d69-9ba9-09ba3acd660f). I think it is a badly written section. I find it difficult to understand and I am a native English speaker. This section is important though:

    <i>In RDP this event is used to synchronize the values of the toggle keys (for example, Caps Lock) **and to
reset the server key state to all keys up.** </i>

   When we receive the event, it is possible that xrdp might think the keys are held down, so we have to send KeyRelease events to the X server. We then toggle keys if necessary so that their state matches the state sent in the event.
3) Scroll lock also needs to be handled somehow.

tsz8899:
@matt335672 Thanks for the technical guidance. I have updated the PR with the following improvements:
  >
  > 1.  Protocol Compliance: Implemented the MS-RDPBCGR 2.2.8.1.1.3.1.1.5 requirement. The synchronization logic now starts by sending KeyRelease events for Caps, Num, and Scroll Lock keys to ensure a
  clean state before alignment.
  > 2.  Full Coverage: Scroll Lock is now included in the synchronization logic.
  > 3.  Accurate State Matching: Replaced the previous comparison with an explicit match between X11 modifier masks (LockMask, Mod2Mask) and the RDP event flags.
  > 4.  Session Reconnect Support: Fixed the persistent state issue by resetting the synchronization flag in rdpLoadLayout, ensuring correct behavior when a user re-connects to an existing session.

matt335672:
Thanks @tsz8899 

There are a lot of tests for specific keycode values (e.g. `keyboard->x11_keycode_caps_lock > 0`). These tests are not needed, as the X server will ignore any key codes which are not valid.

Removing these tests means that `KbdAddEvent()` ends up looking much the same as it did before, except you've made some readability improvements. I think these are worth having. These two sections of code do the same thing, but it's not immediately obvious why - the second is much clearer.

```c
if ((!(xkb_state & LockMask)) != (!(param1 & TS_SYNC_CAPS_LOCK)))
{
    . . .
```

```c
target = (param1 & TS_SYNC_CAPS_LOCK) ? 1 : 0;
current = (xkb_state & LockMask) ? 1 : 0;
if (current != target)
{
    . . .
```

PR #436, opened 2026-09-21
xrdpkeyb: report the session's lock keys to xrdp

MAButz:
The synchronisation of caps, num and scroll lock is one way today: the client reports its state in a Synchronize event when its window takes focus, `KbdSync()` applies it to the session, and nothing ever travels back. Anything that moves the lock state inside the session — an application calling `XkbLockModifiers`, an on screen keyboard, a startup script, or the user simply pressing the key — leaves the lamps on the keyboard in front of the user disagreeing with the session for the rest of it.

This reports the state to xrdp as order 67 whenever the X server's keyboard control changes. It needs neutrinolabs/xrdp#3883 to reach the client; against an older xrdp the order is simply ignored by `lib_mod_process_orders()`.

Two things keep it from feeding back on itself:

- The state the client reported is recorded as reported rather than applied and sent back, on both sides of the `KbdSync()` call — the key events it queues may be handled inside it or after it.
- Nothing is reported before the deferred sync from #419 has run. That also covers a keymap load, since `rdpLoadLayout()` already clears the flag. Until then the session is still catching up with the client, and a data PDU sent while the client is renegotiating makes mstsc.exe hang up. Building on that existing flag is why this change is as small as it is.

Only changes are reported: there is no capability to negotiate for the PDU that carries them downstream.

Tested on Debian 13 against mstsc.exe (Windows 11) and xfreerdp 3.31, both over H.264/GFX.


MAButz:
A note on the red check, so nobody has to dig: **`Build with latest xorg` fails for a reason unrelated to this change.** It fails at step 10, `./configure`, before a line of this patch is compiled. The other eleven checks pass, including `With glamor` on both gcc and clang.

```
checking for xorg-server >= 0... no
configure: error: please install xserver-xorg-dev, xorg-x11-server-sdk or xorg-x11-server-devel
```

`scripts/build_latest_xorg_apt.sh` builds the soft-link that redirects the Debian multiarch pkgconfig directory back to `lib/pkgconfig` from `$(uname -p)`, which on Linux is the processor type and prints `unknown`. It should be `$(uname -m)`, `x86_64`. So the link lands in `lib/unknown-linux-gnu/pkgconfig` while meson installs `xorg-server.pc` into `lib/x86_64-linux-gnu/pkgconfig`, which is not on the job's `PKG_CONFIG_PATH`.

It has been green until now because the script returns early when its target exists, so a cached `xorg-build` never needs the link. The cache expired — the last run of this job was 2026-07-17 — and this PR happens to be the first since to rebuild xorg from scratch:

```
Cache not found for input keys: ubuntu24-cache-xorg-build
```

#437 has the one-word fix. This PR does not touch `configure.ac`, the scripts or the workflows.

MAButz:
A correction to my note above: I wrote that "#437 has the one-word fix". That was wrong, and the CI on #437 itself showed it — the `uname -p` bug is real but is not the cause. #437 now takes a different route, naming the multiarch directory in `PKG_CONFIG_PATH` instead of relying on the soft-link, and is a draft until its own run is green.

The part of the note that stands is the diagnosis for **this** PR: `Build with latest xorg` fails at step 10, `./configure`, before a line of this change is compiled, and it fails the same way on a branch that only touches the CI. The other eleven checks here pass, `With glamor` among them.

MAButz:
Closing the loop on the red check here: #437 is green now and explains it in full. Two things were wrong in the job, both unrelated to this change — the multiarch pkgconfig link was created from `$(uname -p)` instead of `$(uname -m)`, and `xorg-build/share/pkgconfig`, where `xorgproto` puts `xproto.pc`, was not on the search path, so pkg-config fell back to the OS copy at 7.0.33 while the xserver wants >= 7.0.34.

This PR will keep showing the same failure until #437 lands, since its branch does not carry those two commits. The other eleven checks here pass.

PR #437, opened 2026-09-22
ci: fix Build with latest xorg - wrong link name and a missing pkgconfig directory

MAButz:
`Build with latest xorg` fails at `./configure`:

```
checking for xorg-server >= 0... no
configure: error: please install xserver-xorg-dev, xorg-x11-server-sdk or xorg-x11-server-devel
```

Two separate things are wrong, and each alone is enough to produce that line. Both are confirmed by a diagnostic run on this branch, not by reading.

**1. The multiarch link is created under the wrong name.** `build_latest_xorg_apt.sh` builds it from `$(uname -p)`, which on Linux is the processor type and prints `unknown`; the directory is named after the machine type, `$(uname -m)`. The link therefore landed in `lib/unknown-linux-gnu/pkgconfig`, and everything meson installed — `pixman-1.pc`, `libdrm.pc`, `xkbfile.pc`, `xorg-server.pc` itself — stayed out of reach of `PKG_CONFIG_PATH`.

**2. `share/pkgconfig` is not on the search path.** `xorgproto` is architecture independent and installs `xproto.pc` into `share/pkgconfig`, which the job never names. Even with the link fixed, pkg-config skipped the freshly built copy and fell back to the one from the OS:

```
libpkgconf/pkg.c:568 [pkgconf_pkg_try_specific_path]: trying path: .../xorg-build/lib/pkgconfig for xproto
libpkgconf/pkg.c:568 [pkgconf_pkg_try_specific_path]: trying path: /usr/lib/x86_64-linux-gnu/pkgconfig for xproto

Package dependency requirement 'xproto >= 7.0.34' could not be satisfied.
Package 'xproto' has version '7.0.33', required version is '>= 7.0.34'
```

So `configure` reported `xorg-server ... no` with `xorg-server.pc` sitting right in front of it — the file was found, its dependency was not.

**Why this has been green until now.** `build_latest_xorg_apt.sh` returns early when `lib/pkgconfig/xorg-server.pc` already exists, so a restored `xorg-build` never exercises any of this. GitHub drops caches after a week without a hit; the last run of this job was 2026-07-17, and the next PR to come along rebuilt xorg from scratch:

```
Cache not found for input keys: ubuntu24-cache-xorg-build
```

With both changes the job passes, `make check` included, and the cache is populated again.

Found while looking into the failure on #436, which is unrelated to the change there. #436 will keep showing the same red check until this lands, since its branch does not carry these two commits.

*Two earlier versions of this PR were wrong — the first blamed the `uname` bug alone, the second added the multiarch directory to `PKG_CONFIG_PATH` and still failed. Both were disproved by CI here rather than by argument, and the comments below record that. The diagnosis above comes from a run that printed what was on disk and what pkg-config made of it.*


MAButz:
Correcting myself: the first version of this PR did **not** fix the failure. I pushed it on the strength of the `uname -p` bug alone without verifying, and CI on this very PR proved it wrong — `Build with latest xorg` failed again at `./configure`, on a run whose `head_sha` is the corrected script. Sorry for the noise.

What I have actually measured since, with meson 1.12 and a prefix laid out the way the script lays it out:

```
$ ln -s ../pkgconfig pfx/lib/x86_64-linux-gnu/pkgconfig   # target does not exist yet
$ meson install -C builddir
FileExistsError: [Errno 17] File exists: .../pfx/lib/x86_64-linux-gnu/pkgconfig
ERROR: Unhandled python OSError.
```

and with the target present beforehand the file does land in `lib/pkgconfig` as intended. So the soft-link approach works only when `lib/pkgconfig` already exists, and nothing creates it.

What I cannot explain from the job logs is why, with `uname -m` in place, meson installed happily into `lib/x86_64-linux-gnu/pkgconfig` instead of failing as above — the script prints nothing at that point. Rather than keep guessing, this PR now names the directory in `PKG_CONFIG_PATH` and stops depending on the link at all:

```yaml
- name: Add the multiarch pkgconfig directory to PKG_CONFIG_PATH
  run: echo "PKG_CONFIG_PATH=$PKG_CONFIG_PATH:${{ github.workspace }}/xorg-build/lib/$(uname -m)-linux-gnu/pkgconfig" >>$GITHUB_ENV
```

The `uname -p` → `uname -m` change stays as a second commit, but as what it is: a latent bug, not the cause.

Marking this a draft until CI here is green, so nobody merges a fix on my word again.

MAButz:
Green, all twelve checks including `Build with latest xorg`, `make check` among them, and `Post Cache xorg-build` succeeded so `ubuntu24-cache-xorg-build` is populated again.

The diagnostic run answered it. `xorg-server.pc` was being found all along once the link had the right name — what failed was its dependency:

```
trying path: .../xorg-build/lib/pkgconfig for xproto
trying path: /usr/lib/x86_64-linux-gnu/pkgconfig for xproto

Package dependency requirement 'xproto >= 7.0.34' could not be satisfied.
Package 'xproto' has version '7.0.33', required version is '>= 7.0.34'
```

`xorgproto` is built by the script — it is in `MODULAR_PKG_LIST` — but being architecture independent it installs `xproto.pc` into `share/pkgconfig`, which the job never named, so pkg-config walked past it to the OS copy.

The temporary diagnostic commit is gone; what is left is the two-line fix. Description updated, out of draft.

My apologies for the two wrong turns before this. Both were pushed on reasoning rather than measurement, and both were disproved by this PR's own CI.

matt335672:
@MAButz - thanks for your contribution!

